PR #63, opened 2014-05-11
Sketch of fix for delegate private method stubbing

fables-tales:
A sketch of a fix for https://github.com/rspec/rspec/issues/42

I paired on this with @bestie. It seems to solve the problem, but I'm really not sure how solid this is a solution. The basic problem here is that when asking a delegator if it has a particular method, it raises a warning if that method is a private method that is defined on `Kernel`. The solution is to check if the delegator object itself implements that method, and if it does, take that implementation. Otherwise we take the implementation off kernel itself. I'm not sure how I feel about this at all, but the fix does seem to deal with the issue.


aledalgrande:
Was this issue ever fixed?

PR #134, opened 2014-11-20
Test EncodedSring#to_s for undefined conversion / invalid byte sequence

bf4:
Related to https://github.com/rspec/rspec-core/pull/1760

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,19 +22,24 @@ rvm:
   - 2.2
   - ruby-head
   - ree
-  - jruby-18mode
-  - jruby
-  - jruby-head
   - rbx
 matrix:
   include:
     - rvm: jruby
-      env: JRUBY_OPTS='--2.0'
-  allow_failures:
+      env: JRUBY_OPTS='--server -Xcompile.invokedynamic=false -Xcompat.version=2.0'
     - rvm: jruby-head
+      env: JRUBY_OPTS='--server -Xcompile.invokedynamic=false'
+    # These two are temporary until https://github.com/travis-ci/travis-ci/issues/3067 is solved.
+    - rvm: jruby-18mode
+      env: JRUBY_OPTS='--server -Xcompile.invokedynamic=false'
+    - rvm: jruby
+      env: JRUBY_OPTS='--server -Xcompile.invokedynamic=false'
+  allow_failures:
     - rvm: ruby-head
     - rvm: rbx
     # These two are temporary until https://github.com/travis-ci/travis-ci/issues/3067 is solved.
     - rvm: jruby-18mode
+      env: JRUBY_OPTS='--server -Xcompile.invokedynamic=false'
     - rvm: jruby
+      env: JRUBY_OPTS='--server -Xcompile.invokedynamic=false'
   fast_finish: true
--- a/Gemfile
+++ b/Gemfile
@@ -14,7 +14,8 @@ branch = File.read(File.expand_path("../maintenance-branch", __FILE__)).chomp
 end
 
 ### dep for ci/coverage
-gem 'simplecov', '~> 0.8'
+gem 'simplecov', '~> 0.9'
+gem 'simplecov-html', :github => 'colszowka/simplecov-html'
 
 gem 'rubocop', "~> 0.23.0", :platform => [:ruby_19, :ruby_20, :ruby_21]
 
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -51,8 +51,6 @@ module RSpec
         finalize_output(output, hunks.last.diff(format_type).to_s) if hunks.last
 
         color_diff output
-      rescue Encoding::CompatibilityError
-        handle_encoding_errors
       end
       # rubocop:enable MethodLength
 
@@ -200,17 +198,6 @@ module RSpec
         end
       else
         def pick_encoding(_source_a, _source_b)
-        end
-      end
-
-      def handle_encoding_errors
-        if @actual.source_encoding != @expected.source_encoding
-          "Could not produce a diff because the encoding of the actual string " \
-          "(#{@actual.source_encoding}) differs from the encoding of the expected " \
-          "string (#{@expected.source_encoding})"
-        else
-          "Could not produce a diff because of the encoding of the string " \
-          "(#{@expected.source_encoding})"
         end
       end
     end
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -190,6 +190,7 @@ module RSpec
       end
 
       if String.method_defined?(:encoding)
+        # see https://github.com/ruby/ruby/blob/ca24e581ba/encoding.c#L1191
         def pick_encoding(source_a, source_b)
           Encoding.compatible?(source_a, source_b) || Encoding.default_external
         end
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -31,7 +31,7 @@ module RSpec
 
       # rubocop:disable MethodLength
       def diff_as_string(actual, expected)
-        @encoding = pick_encoding actual, expected
+        @encoding = EncodedString.pick_encoding(actual, expected)
         @actual   = EncodedString.new(actual, @encoding)
         @expected = EncodedString.new(expected, @encoding)
         output    = EncodedString.new("\n", @encoding)
@@ -188,16 +188,6 @@ module RSpec
           object =~ /\n/ ? object : object.inspect
         else
           PP.pp(object, "")
-        end
-      end
-
-      if String.method_defined?(:encoding)
-        # see https://github.com/ruby/ruby/blob/ca24e581ba/encoding.c#L1191
-        def pick_encoding(source_a, source_b)
-          Encoding.compatible?(source_a, source_b) || Encoding.default_external
-        end
-      else
-        def pick_encoding(_source_a, _source_b)
         end
       end
     end
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -7,8 +7,14 @@ module RSpec
   module Support
     # rubocop:disable ClassLength
     class Differ
+      if String.method_defined?(:encoding)
+        EMPTY_DIFF = EncodedString.new("", Encoding.default_external)
+      else
+        EMPTY_DIFF = EncodedString.new("")
+      end
+
       def diff(actual, expected)
-        diff = ""
+        diff = EMPTY_DIFF.dup
 
         if actual && expected
           if all_strings?(actual, expected)
@@ -26,11 +32,9 @@ module RSpec
       # rubocop:disable MethodLength
       def diff_as_string(actual, expected)
         @encoding = pick_encoding actual, expected
-
         @actual   = EncodedString.new(actual, @encoding)
         @expected = EncodedString.new(expected, @encoding)
-
-        output = EncodedString.new("\n", @encoding)
+        output    = EncodedString.new("\n", @encoding)
 
         hunks.each_cons(2) do |prev_hunk, current_hunk|
           begin
--- a/lib/rspec/support/encoded_string.rb
+++ b/lib/rspec/support/encoded_string.rb
@@ -2,6 +2,16 @@ module RSpec
   module Support
     # @private
     class EncodedString
+      if String.method_defined?(:encoding)
+        # see https://github.com/ruby/ruby/blob/ca24e581ba/encoding.c#L1191
+        def self.pick_encoding(source_a, source_b)
+          Encoding.compatible?(source_a, source_b) || Encoding.default_external
+        end
+      else
+        def self.pick_encoding(_source_a, _source_b)
+        end
+      end
+
       # Ruby's default replacement string for is U+FFFD ("\xEF\xBF\xBD") for Unicode encoding forms
       #   else is '?' ("\x3F")
       REPLACE = "\x3F"
--- a/lib/rspec/support/encoded_string.rb
+++ b/lib/rspec/support/encoded_string.rb
@@ -81,7 +81,6 @@ module RSpec
         end
         # End: Needed for 1.9.2
 
-
         def detect_source_encoding(string)
           string.encoding
         end
--- a/lib/rspec/support/encoded_string.rb
+++ b/lib/rspec/support/encoded_string.rb
@@ -78,16 +78,17 @@ module RSpec
         # Encoding::CompatibilityError
         #
         def matching_encoding(string)
+          encoding = EncodedString.pick_encoding(source_encoding, @encoding)
           # Converting it to a higher character set (UTF-16) and then back (to UTF-8)
           # ensures that we strip away invalid or undefined byte sequences
           # => no need to rescue Encoding::InvalidByteSequenceError, ArgumentError
           string.encode(::Encoding::UTF_16LE, ENCODING_STRATEGY[:bad_bytes]).
-            encode(@encoding)
+            encode(encoding)
         rescue Encoding::UndefinedConversionError, Encoding::CompatibilityError
-          string.encode(@encoding, ENCODING_STRATEGY[:cannot_convert])
+          string.encode(encoding, ENCODING_STRATEGY[:cannot_convert])
         # Begin: Needed for 1.9.2
         rescue Encoding::ConverterNotFoundError
-          string.force_encoding(@encoding).encode(ENCODING_STRATEGY[:no_converter])
+          string.force_encoding(encoding).encode(ENCODING_STRATEGY[:no_converter])
         end
         # End: Needed for 1.9.2
 
--- a/lib/rspec/support/encoded_string.rb
+++ b/lib/rspec/support/encoded_string.rb
@@ -4,7 +4,6 @@ module RSpec
     class EncodedString
       # Ruby's default replacement string for is U+FFFD ("\xEF\xBF\xBD") for Unicode encoding forms
       #   else is '?' ("\x3F")
-      MRI_UNICODE_UNKOWN_CHARACTER = "\xEF\xBF\xBD"
       REPLACE = "\x3F"
 
       def initialize(string, encoding=nil)
@@ -36,6 +35,24 @@ module RSpec
 
         private
 
+        ENCODING_STRATEGY = {
+          :bad_bytes => {
+            :invalid => :replace,
+            # :undef   => :nil,
+            :replace => REPLACE
+          },
+          :cannot_convert => {
+            # :invalid => :nil,
+            :undef   => :replace,
+            :replace => REPLACE
+          },
+          :no_converter => {
+            :invalid => :replace,
+            # :undef   => :nil,
+            :replace => REPLACE
+          }
+        }
+
         # Raised by Encoding and String methods:
         #   Encoding::UndefinedConversionError:
         #     when a transcoding operation fails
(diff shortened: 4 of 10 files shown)


lib/rspec/support/differ.rb
```diff
@@ -47,8 +47,6 @@ def diff_as_string(actual, expected)
         finalize_output(output, hunks.last.diff(format_type).to_s) if hunks.last
 
         color_diff output
-      rescue Encoding::CompatibilityError
-        handle_encoding_errors
```
JonRowe: Given how notoriously difficult it is to actually craft an invalidly encoded string when you run a conventually encoded environment I'm uncomfortable with this FYI

bf4:
Hammering away at failures in 1.9.2 and discovered that some of the failure reporting is incorrect due to.. wait for it, using the differ to test the differ.

For example, when we say `actual = differ.diff(str1, str2); expect(actual).to eq(expected)`, what we are really saying is: `actual = differ.diff(str1, str2); actual != expected && RSpec::Expectations.fail_with(message, actual, expected)`.

`RSpec::Expectations.fail_with` calls `differ.diff(actual, expected)` where `differ` is `**RSpec::Support::Differ.new**(:object_preparer => lambda { |object| RSpec::Matchers::Composable.surface_descriptions_in(object) }, :color => RSpec::Matchers.configuration.color?)`
 

So, all these differ_spec tests where we say `expect(str1).to eq(str2)` will run through the differ twice on failure.  We should be testing them as `expect(str1 == str2).to eq(true)`

bf4:
Signing off till Sunday, will be back. Am playing with ideas for a different method of string expectation for the differ

```ruby
      SAFE_CHR = Hash.new { |h, x| h[x] = x.chr rescue ("U+%.4X=" % [x]) }
      def safe_codepoints(str)
        str.each_codepoint.map {|codepoint| SAFE_CHR[codepoint] }
      rescue ArgumentError
        str.each_byte.map {|byte| SAFE_CHR[byte] }
      end
      def expect_identical_string(str1, str2, expected_encoding = str1.encoding)
        expect(str1.encoding).to eq(expected_encoding)

        str1_bytes = safe_codepoints(str1)
        str2_bytes = safe_codepoints(str2)
        return unless str1_bytes != str2_bytes
        str1_differences = []
        str2_differences = []
        str2_bytes.each_with_index do |str2_byte, index|
          str1_byte = str1_bytes.fetch(index) do
            str2_differences.concat str2_bytes[index..-1]
            return
          end
          if str1_byte != str2_byte
            str1_differences << str1_byte
            str2_differences << str2_byte
          end
        end
        expect(str1_differences.join).to eq(str2_differences.join)
      end

```

bf4:
I expect there to be a bit of cleanup as it's been a lot of work and I'm tired, but I think it's about done.

I'll look at the failures

```plain
JRuby


 RSpec::Support::EncodedString#<< with two ascii strings with a target encoding of UTF-8  has an encoding of UTF-8
     Failure/Error: expect(resulting_string.encoding.to_s).to eq('UTF-8')
       
       expected: "UTF-8"
            got: "ASCII-8BIT"

rbx
 1) RSpec::Support::EncodedString#split when splitting a string that is not US-ASCII compatible normally raises an Encoding::CompatibilityError
     Failure/Error: expect {
       expected Encoding::CompatibilityError but nothing was raised
rspec ./spec/rspec/support/encoded_string_spec.rb:201 #
```

bf4:
urgh

I think one of these failures is due to https://github.com/colszowka/simplecov-html/pull/32 not being merged and released (failures in coverage subprocess).  

`ruby -E` doesn't work on 1.8.7, so maybe I should just go back to using env variables. I didn't properly test that.

Some of the encoding expectations are a little too aggressive, I think.  I should only 'enforce' internal/external and just 'skip/document' locale/filesystem, as they do have effects.

bf4:
Ok, now it looks good except for some unrelated failures.  

Obviously the changes to script/functions.sh and .travis.yml would have to be pushed to rspec-dev, if the look good to you.

For shelling out I've enjoyed using this snippet as it lets me stream the output as the process is running, rather than having to wait.

```ruby
require 'pty'
# should consider trapping SIGINT in here
def run(cmd)
  puts cmd
  child_process = ''
  result = ''
  # http://stackoverflow.com/a/1162850
  # stream output of subprocess
  begin
    PTY.spawn( cmd ) do |stdin, stdout, pid|
      begin
        # Do stuff with the output here. Just printing to show it works
        stdin.each do |line|
          print line
          result << line
        end
        child_process = PTY.check(pid)
      rescue Errno::EIO
        puts "Errno:EIO error, but this probably just means " +
              "that the process has finished giving output"
      end
    end
  rescue PTY::ChildExited
    puts "The child process exited!"
  end
  unless child_process.success?
    exitstatus = child_process.exitstatus
    puts "FAILED: #{child_process.pid} exited with status #{exitstatus.inspect} due to failed command #{cmd}"
    exit exitstatus || 1
  end
  result
end
```

though in [metric_fu I do use Open3.popen3](https://github.com/metricfu/metric_fu/blob/1adc860c6573086c93b0b99b0fa04bbc9b847c88/lib/metric_fu/gem_run.rb)

```ruby
    def execute
      mf_debug "Running #{summary}"
      captured_output = ''
      captured_errors = ''
      thread = ''
      Open3.popen3("#{library_name}", *arguments) do |stdin, stdout, stderr, wait_thr|
        captured_output << stdout.read.chomp
        captured_errors << stderr.read.chomp
        thread = wait_thr
      end
    rescue StandardError => run_error
      handle_run_error(run_error)
    rescue SystemExit => system_exit
      handle_system_exit(system_exit)
    ensure
      print_errors
      return captured_output, captured_errors, thread.value
    end

    def handle_run_error(run_error)
      @errors << "ERROR: #{run_error.inspect}"
    end

    def handle_system_exit(system_exit)
      status =  system_exit.success? ? "SUCCESS" : "FAILURE"
      message = "#{status} with code #{system_exit.status}: " <<
        "#{system_exit.message}: #{system_exit.backtrace.inspect}"
      if status == 'SUCCESS'
        mf_debug message
      else
        @errors << message
      end
    end

    def print_errors
      return if @errors.empty?
      STDERR.puts "ERRORS running #{summary}"
      @errors.each do |error|
        STDERR.puts "\t" << error
      end
    end
```

bf4:
One other question is if you'd like to use the fallback option for undefined conversions so that an e.g. '?' will be instead replaced by 'U+FFFD'.

e.g.
```ruby
module Support::EncodedString
  if String.method_defined?(:encoding)
     # REPLACE =
+     UNDEF_FALLBACK = Hash.new { |h, x| h[x] = "U+%.4X=" % [x] }
      # snip ENCODING_STRATEGY =
        :cannot_convert => {
            # :invalid => :nil,                                                                                                                                                                           
            :undef   => :replace,
-           :replace => REPLACE,
+            :fallback => UNDEF_FALLBACK
          },  
```

lib/rspec/support/encoded_string.rb
```diff
@@ -2,7 +2,19 @@ module RSpec
...
+          Encoding.compatible?(source_a, source_b) || Encoding.default_external
+        end
+      else
+        def self.pick_encoding(_source_a, _source_b)
+        end
+      end
+
+      # Ruby's default replacement string for is U+FFFD ("\xEF\xBF\xBD") for Unicode encoding forms
+      #   else is '?' ("\x3F")
+      REPLACE = "\x3F"
```
myronmarston: Can this just be `REPLACE = "?"`?  I find the `\x` escaped form to be much harder to read (and it winds up requiring a comment to explain it's a question mark).

Also why is this better than the default replacement string of `U+FFFD`?  I don't understand that from your comment.

lib/rspec/support/encoded_string.rb
```diff
@@ -2,7 +2,19 @@ module RSpec
...
+        # see https://github.com/ruby/ruby/blob/ca24e581ba/encoding.c#L1191
+        def self.pick_encoding(source_a, source_b)
+          Encoding.compatible?(source_a, source_b) || Encoding.default_external
+        end
+      else
+        def self.pick_encoding(_source_a, _source_b)
+        end
+      end
+
+      # Ruby's default replacement string for is U+FFFD ("\xEF\xBF\xBD") for Unicode encoding forms
```
myronmarston: This sentence doesn't make grammatical sense ("string for is" reads awkwardly and I'm not sure what you're trying to say).  Can you rephrase it?

Gemfile
```diff
@@ -14,7 +14,8 @@ branch = File.read(File.expand_path("../maintenance-branch", __FILE__)).chomp
 end
 
 ### dep for ci/coverage
-gem 'simplecov', '~> 0.8'
+gem 'simplecov', '~> 0.9'
+gem 'simplecov-html', :github => 'colszowka/simplecov-html'
```
myronmarston: One of your comments mentioned this addresses a failure but I don't understand...we didn't need this gem before so why do we need it now? What failure does it address?

.travis.yml
```diff
@@ -22,19 +22,24 @@ rvm:
...
-      env: JRUBY_OPTS='--2.0'
-  allow_failures:
+      env: JRUBY_OPTS='--server -Xcompile.invokedynamic=false -Xcompat.version=2.0'
     - rvm: jruby-head
+      env: JRUBY_OPTS='--server -Xcompile.invokedynamic=false'
+    # These two are temporary until https://github.com/travis-ci/travis-ci/issues/3067 is solved.
+    - rvm: jruby-18mode
+      env: JRUBY_OPTS='--server -Xcompile.invokedynamic=false'
+    - rvm: jruby
+      env: JRUBY_OPTS='--server -Xcompile.invokedynamic=false'
```
myronmarston: Why all these JRUBY_OPTS changes?  I don't understand what these options do and I'd prefer to keep our JRUBY_OPTS simpler without a clear reason to add the complication of these options...especially since we haven't needed these options in the other repos or in this repo before now.

lib/rspec/support/differ.rb
```diff
@@ -7,8 +7,14 @@ module RSpec
...
     class Differ
+      if String.method_defined?(:encoding)
+        EMPTY_DIFF = EncodedString.new("", Encoding.default_external)
+      else
+        EMPTY_DIFF = EncodedString.new("")
+      end
+
       def diff(actual, expected)
-        diff = ""
+        diff = EMPTY_DIFF.dup
```
myronmarston: Why was `""` insufficient?

Also, is the `dup` just to guard against mutations (to ensure `EMPTY_DIFF` isn't mutated)?  If there aren't any mutations maybe you could freeze the constant instead and use it w/o duping for fewer allocations.

lib/rspec/support/encoded_string.rb
```diff
@@ -33,21 +45,52 @@ def to_s
...
+            # :invalid => :nil,
+            :undef   => :replace,
+            :replace => REPLACE
+          },
+          :no_converter => {
+            :invalid => :replace,
+            # :undef   => :nil,
+            :replace => REPLACE
+          }
+        }
```
myronmarston: What's with the commented out entries in these hashes?

myronmarston: Also, is there any real benefit to sticking all these strategies in a big named hash?  Why not just make them separate constants and refer to them separately?  That would remove an unnecessary method call (`Hash#[]`) and it seems like you're just using the hash as an alternate form of named constants.

lib/rspec/support/encoded_string.rb
```diff
@@ -33,21 +45,52 @@ def to_s
...
+
+        # Raised by Encoding and String methods:
+        #   Encoding::UndefinedConversionError:
+        #     when a transcoding operation fails
+        #     e.g. "\x80".encode('utf-8','ASCII-8BIT')
+        #   Encoding::InvalidByteSequenceError:
+        #     when the string being transcoded contains a byte invalid for the either
+        #     the source or target encoding
```
myronmarston: Should "...for the either the source or target..." be "...for either the source or target..."?

lib/rspec/support/spec/encoding_helpers.rb
```diff
@@ -0,0 +1,61 @@
...
+  module Support
+    module EncodingHelpers
+      module_function
+
+      # For undefined conversions, replace as "U+<codepoint>"
+      # e.g. '\xa0' becomes 'U+00A0'
+      # see https://github.com/ruby/ruby/blob/34fbf57aaa/test/ruby/test_transcode.rb#L2050
+      def safe_chr
+        # rubocop:disable Style/RescueModifier
+        @safe_chr ||= Hash.new { |h, x| h[x] = x.chr rescue ("U+%.4X" % [x]) }
```
myronmarston: I consider trailing `rescue` to be an anti-pattern since it's such a wide `rescue`.  Can this rescue the specific exception class you are handling here instead?

lib/rspec/support/spec/encoding_helpers.rb
```diff
@@ -0,0 +1,61 @@
...
+        # rubocop:enable Style/RescueModifier
+      end
+
+      if String.method_defined?(:encoding)
+
+        def safe_codepoints(str)
+          str.each_codepoint.map { |codepoint| safe_chr[codepoint] }
+        rescue ArgumentError
```
myronmarston: What part of `str.each_codepoint.map { |codepoint| safe_chr[codepoint] }` is capable of raising an `ArgumentError`?

lib/rspec/support/spec/encoding_helpers.rb
```diff
@@ -0,0 +1,61 @@
...
+              return
+            end
+            if str1_byte != str2_byte
+              str1_differences << str1_byte
+              str2_differences << str2_byte
+            end
+          end
+          # rubocop:enable Style/Next
+          expect(str1_differences.join).to eq(str2_differences.join)
+        end
```
myronmarston: You mentioned before that `expect(str1 == str2).to eq true` would bypass the differ (and you're right) so why is that insufficient?  What does this far more complex way of comparing two strings gain you?

script/functions.sh
```diff
@@ -112,7 +116,7 @@ function check_documentation_coverage {
 }
 
 function check_style_and_lint {
-  echo "bin/rubucop lib"
+  echo "bin/rubocop lib"
```
myronmarston: Ha, good catch!

script/functions.sh
```diff
@@ -7,6 +7,10 @@ source $SCRIPT_DIR/predicate_functions.sh
 
 # idea taken from: http://blog.headius.com/2010/03/jruby-startup-time-tips.html
 export JRUBY_OPTS="${JRUBY_OPTS} -X-C" # disable JIT since these processes are so short lived
+# Set the external encoding to UTF-8 in a 1.8.7-compatible way
+export LANG=en_US.UTF-8
+export LC_ALL=en_US.UTF-8
```
myronmarston: Why is this needed now when it wasn't before? I thought the external encoding was already set to UTF-8?

myronmarston:
Hey @bf4, I really appreciate the effort you're putting into this, but I also feel like this PR keeps shifting significantly every time I take a look at it.  As lead RSpec maintainer I think it's important that I understand any code I merge (as I'll need to be able to refactor it and fix bugs in it in the future) but there's a lot going on here that I don't understand and every time I look at it it seems like there's new stuff I don't understand and I still don't understand the old stuff :(.  At this point it seems like here are many concerns being addressed all in one PR:

* Improving the existing specs.
* Improving the build.
* Handling some encoding edge cases that we didn't already handle.

Having all these concerns is making it much harder for me to understand what the actual affect of this PR is and also to ensure we're not losing coverage in all the spec changes.  I hate making more work for you (especially since you've done so much work on this PR already...) but do you think you could open up a few smaller PRs that address a single aspect each?  That would make your changes much easier to review and reason about and hopefully we could get the stuff that's not still an open question merged quickly and then be left with whatever the WIP stuff is in a much smaller PR.

Ideally I'd like to keep spec refactorings distinct from implementation refactorings and bug fixes.

myronmarston:
> For shelling out I've enjoyed using this snippet as it lets me stream the output as the process is running, rather than having to wait.

What does that have to do with your changes here?  (I don't see any relation...) Or are you just recommending we change how rspec-support handles shelling out? If it's the latter, please open a separate PR for that -- mixing concerns all in one PR is making my head spin!

bf4:
@myronmarston thanks for your review.  I think you're right that this can now be broken up into smaller PRs.  As you know, that happens sometimes that some changes expose others :)

I know the scope has changed a bit, but that's mostly because I've been trying to figure out how to 1) cover the different Encoding exceptions and 2) reliably test them.

I see a few types of comments

1. About env variables JRUBY_OPTS and LC_etc.
2. Typos or otherwise unclear comments
3. Code conventions / practices: e.g. constant lookup, allocations, hash, etc
4. Understanding the exceptions rescue or other defensive techniques

 I'll answer in place, later, I think, but for now:

1. JRUBY_OPTS are just a workaround for travis failures regarding the removed cext option causing warnings which cause failures https://github.com/rspec/rspec-support/commit/8aa73ae65d65e535d4d16df63fb9bdd51a94a37e.  The LC_etc changes are to make explicit that we expect our tests to run with external UTF-8 encoding.  Travis does this, but our suite does not.  So, on my vagrant box (rake-dev-compiler, btw), the default encoding is ISO-8859-1, which causes expectation failures.  Rather than debug expectation failures stemming from different external encodings, I just made sure we always test against the same external encoding.  https://github.com/bf4/rspec-support/commit/db2c3a43e1cdb0fc1491394328f78c712aa9ed19  In the future, we might want to try something like [`EnvUtil.with_default_external`](https://github.com/jruby/jruby/blob/c1be61a501d1295fa1ea9f894e1a8e186411f32a/test/mri/ruby/envutil.rb#L150) to test these cases explicitly.
2. Typos.. yeah, sorry, happens
3. I'll change per your recommendations, which I agree with.
4. The safe_chr rescue is a RangeError.  If the codepoint is `128169` (pile of poo), then `128169.chr` raises `RangeError: 128169 out of char range` in which case it returns ` ("U+%.4X" % [128169]) #=> "U+1F4A9"`.  Contrariwise, `63.chr` returns `?`.  The idea behind the safe_chr, safe_codepoints, and expect_identical_string is to provide the most useful failure info if the strings aren't the same.

First, it checks if the strings are the same encoding. If not, it's a failure.
Next, it turns them into bytes/codepoints for comparison. If they're the same, success!  If not, we try to produce the most useful failure message.

If we did a straight-up bytes comparison, we would just get a diff of numbers which is pretty undreadable.  So, what I try to do, is encode all the codepoints I can, then compare the characters in the strings until there is a difference, then output that.

Rescuing the ArgumentError in safe_codepoints is if the string has an invalid byte sequence, it raises that on string.codepoints, so I fall back to just bytes.

There's probably other ways to do it :)

myronmarston:
> JRUBY_OPTS are just a workaround for travis failures regarding the removed cext option causing warnings which cause failures 8aa73ae.

From what I can tell, it looks like that didn't actually fix the JRuby build?  According to the travis issue there was an RVM bug that caused jruby-head to get installed as jruby so we've just been planning to wait until the fix is rolled out.  If there was a simple fix like that I'd be in favor but I'd want it made in rspec-dev and pushed out to all rspec repos.

> The LC_etc changes are to make explicit that we expect our tests to run with external UTF-8 encoding. Travis does this, but our suite does not. So, on my vagrant box (rake-dev-compiler, btw), the default encoding is ISO-8859-1, which causes expectation failures. Rather than debug expectation failures stemming from different external encodings, I just made sure we always test against the same external encoding. bf4@db2c3a4 In the future, we might want to try something like EnvUtil.with_default_external to test these cases explicitly.

I think that `EnvUtil.with_default_external` might be a better solution, given that users who don't have `UTF-8` encoding probably won't run the travis scripts to run the specs.  It also makes it explicit that the encoding is being set for particular specs that rely on it.  Lastly, that helper will allow us to write specs for environments that have a different default external coding if we want.

>  The idea behind the safe_chr, safe_codepoints, and expect_identical_string is to provide the most useful failure info if the strings aren't the same.

Gotcha.  Thanks for explaining that.  I think the logic you've written could benefit from being refactored into a custom matcher, that uses a simple string equality check for `matches?`, is diffable, and exposes the `safe_chr`/`safe_codepoints` values from `failure_message`, `expected` and `actual` -- that way they will get used in the failure diff.  The nice thing about splitting the logic that way is that it keeps the core matching logic simple and easy to reason about, and does the more complicated conversion only when necessary to provide the most useful failure message.  If you want help creating such a matcher, let me know.

lib/rspec/support/spec/encoding_helpers.rb
bf4: I was actually thinking these helpers need their own tests :)

bf4:
Let me know what you'd like for next steps here. What's remaining after #151 and #152 

1. [Handling the ArgumentError](https://github.com/rspec/rspec-support/commit/5c54a6e9a88f7d62d5759b41c4eba62ea439a4f0#diff-6f77530d5756f8c02ca078ddecaa891cR71)
2. [Adding Differ specs, have differ empty string be an EncodedString](https://github.com/rspec/rspec-support/commit/2615f3d1be63e128a3a2f962cd18966a101fd5df)
3. [Differ no longer needs to rescue CompatibilityErrors](https://github.com/rspec/rspec-support/commit/d431a4d7a6981cd80fdbf72b4ce6d73b26a1bdf3)
3. [Making use of getting the compatible encoding before converting, as `Differ#pick_encoding`](https://github.com/rspec/rspec-support/commit/93776b9ccc6e613be82d32dec82a3f07fae7794f)
4. [Adding some way to testing for an external encoding like `EnvUtil.with_default_external`, esp. for tests that only pass when it's UTF-8.](https://github.com/rspec/rspec-support/pull/134#issuecomment-68937671)
5. Improving failure messages via the matcher and/or changing the replace string and/or [using a fallback method.](https://github.com/rspec/rspec-support/pull/134#issuecomment-68831901)

bf4:
Looks like the big goals of this [1 through 4](https://github.com/rspec/rspec-support/pull/134#issuecomment-68984440) are done.  The only things left include 

> 5. Adding some way to testing for an external encoding like `EnvUtil.with_default_external`, esp. for tests that only pass when it's UTF-8.
> 6. Improving failure messages via the matcher and/or changing the replace string and/or using a fallback method.

And with that, I think we can close this PR.

PR #147, opened 2015-01-01
This spec has been intermittently failing on ruby 2.2 due to `birthtime`

myronmarston:
Locally I can’t repro, unfortunately :(.

### DIFF
--- a/spec/rspec/support/spec/stderr_splitter_spec.rb
+++ b/spec/rspec/support/spec/stderr_splitter_spec.rb
@@ -22,6 +22,9 @@ describe 'RSpec::Support::StdErrSplitter' do
   it 'conforms to the stderr interface' do
     stderr_methods = stderr.methods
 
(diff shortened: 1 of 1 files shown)


myronmarston:
Merging since we've been getting intermittent failures.  Someone can still review this if you like and we can address feedback in a follow up PR as needed.

cupakromer:
I think I may know what's going on here. However, even if it stand error does, sometimes, respond to `:birthtime` why would that make the spec fail? Doesn't the implementation delegate behind the scenes to it when `respond_to?` is called?

myronmarston:
> I think I may know what's going on here. However, even if it stand error does, sometimes, respond to `:birthtime` why would that make the spec fail? Doesn't the implementation delegate behind the scenes to it when `respond_to?` is called?

Yep, that's how it's implemented, so I'm puzzled by the failure.  It suggests that in some circumstances, `$stderr.methods` returns an array including `:birthtime` but `$stderr.respond_to?(:birthtime)` returns false.  No idea why.

PR #148, opened 2015-01-01
Another attempt at travis bundler caching.

myronmarston:
For rspec/rspec-dev#101. The changes in #102 don’t seem
to be caching properly. I’m switching back to a single
shared bundle directory but manually caching it rather
than using `cache: bundler` so that `bundle clean` doesn’t
run.

I plan to put this in rspec-dev (and all repos) if it works
but wanted to test on just one repo for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,17 @@
-# This file was generated on 2014-12-31T23:38:57-08:00 from the rspec-dev repo.
+# This file was generated on 2014-12-29T15:16:37-08:00 from the rspec-dev repo.
 # DO NOT modify it by hand as your changes will get lost the next time it is generated.
 
 language: ruby
 sudo: false
 cache:
   directories:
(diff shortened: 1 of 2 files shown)


myronmarston:
This appears to be working.  Compare this build:

https://travis-ci.org/rspec/rspec-support/builds/45625577

...to this one:

https://travis-ci.org/rspec/rspec-support/builds/45629964

The first build took 37 commutative minutes (3-5 minutes per matrix entry).  The latter took 11.5 minutes cumulatively (with all builds under 90 seconds).  That's a huge improvement.

The one problem is that it doesn't clean the bundle, so the bundle will grow infinitely as new gem versions are released and get installed in our bundles.  I'm not sure what to do about that yet.  I'll go open a PR in rspec-dev with these changes.

myronmarston:
Closing in favor of #149.

PR #151, opened 2015-01-07
Extracting EncodedString specs from #134

bf4:
See https://github.com/rspec/rspec-support/pull/134

I need help with the matcher

Also, I decided not to include the new differ specs in here,
to reduce complexity, but am open to adding it.


lib/rspec/support/encoded_string.rb
```diff
@@ -33,17 +49,55 @@ def to_s
...
+        # Raised by transcoding methods:
+        #   Encoding::ConverterNotFoundError:
+        #     when a named encoding does not correspond with a known converter
+        #     e.g. 'abc'.force_encoding('UTF-8').encode('foo')
+        #     or a converter path cannot be found
+        #     e.g. "\x80".force_encoding('ASCII-8BIT').encode('Emacs-Mule')
+        #
+        # Raised by byte <-> char conversions
+        #  RangeError: out of char range
+        #   e.g. the UTF-16LE emoji: 128169.chr
```
bf4: Is this too much detail in the comments?

lib/rspec/support/spec/encoding_helpers.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+end
+require 'rspec/expectations'
+# Special matchers for comparing encoded strings so that
+# we don't run any expectation failures through the Differ,
+# which also relies on EncodedString. Instead, confirm the
+# strings have the same encoding and same bytes.
+RSpec::Matchers.define :be_identical_string do |expected|
+  match do |actual|
+    RSpec::Support::EncodingHelpers.expect_identical_string(actual, expected)
```
bf4: I know this isn't really right, and now it no longer checks the optional expected encoding, but I thought I'd stub it out for discussion. Also added a comment as to why it's there.  I admit, though, my words confuse characters, codepoints, and bytes.

 see the encoding check as a pre-req for the byte check. I'm not sure if it should be in the same expectation, a separate one, or chained.

lib/rspec/support/encoded_string.rb
```diff
@@ -33,17 +49,55 @@ def to_s
...
         rescue Encoding::ConverterNotFoundError
-          normalize_missing(string.force_encoding(@encoding).encode(:invalid => :replace))
+          normalize_missing(string.force_encoding(@encoding).encode(ENCODE_NO_CONVERTER))
         end
 
         def normalize_missing(string)
-          if @encoding.to_s == "UTF-8"
-            string.gsub(MRI_UNICODE_UNKOWN_CHARACTER.force_encoding(@encoding), "?")
+          if @encoding.to_s == UTF_8
+            string.gsub(MRI_UNICODE_UNKOWN_CHARACTER.force_encoding(@encoding), REPLACE)
```
bf4: There're no functional changes in this file.  Just more tests.

spec/rspec/support/encoded_string_spec.rb
```diff
@@ -54,23 +151,72 @@ module RSpec::Support
...
+          let(:message_with_invalid_byte_sequence) { "\xEF \255 \xAD I have bad bytes".force_encoding(utf8_encoding) }
+
+          it 'normally raises an ArgumentError' do
+            expect {
+              message_with_invalid_byte_sequence.split("\n")
+            }.to raise_error(ArgumentError)
+          end
+
+          it 'replaces invalid bytes with the REPLACE string' do
+            skip 'test is raising an ArgumentError on split'
```
bf4: Here's the failing test I fixed in https://github.com/bf4/rspec-support/commit/5c54a6e9a88f7d62d5759b41c4eba62ea439a4f0#diff-6f77530d5756f8c02ca078ddecaa891cR71

spec/rspec/support/encoded_string_spec.rb
```diff
@@ -54,23 +151,72 @@ module RSpec::Support
...
+            expect {
+              non_ascii_compatible_string.split("\n")
+            }.to raise_error(Encoding::CompatibilityError)
+          end
+
+          it 'makes no changes to the resulting string' do
+            resulting_array = build_encoded_string(non_ascii_compatible_string).split("\n")
+            expect(resulting_array.size).to eq(1) # sanity check
+            expected_string = non_ascii_compatible_string
+            expect_identical_string(resulting_array.first, expected_string, Encoding.find('UTF-16LE'))
```
bf4: See matcher comment. The `Encoding.find('UTF-16LE')` isn't used here, but I want it to be, somehow.

lib/rspec/support/spec.rb
```diff
@@ -12,6 +13,7 @@
   c.include RSpecHelpers
   c.include RSpec::Support::WithIsolatedStdErr
   c.include RSpec::Support::FormattingSupport
+  c.include RSpec::Support::EncodingHelpers
```
bf4: Probably not necessary with a matcher, but wanted to keep it for discussion.

spec/rspec/support/encoded_string_spec.rb
```diff
@@ -1,51 +1,148 @@
+# encoding: utf-8
 require 'spec_helper'
 require 'rspec/support/encoded_string'
 
 module RSpec::Support
   describe EncodedString do
-    let(:target_encoding) { 'UTF-8' }
+    let(:utf8_encoding) { 'UTF-8' }
```
bf4: I think this just clarifies the language

lib/rspec/support/encoded_string.rb
```diff
@@ -33,17 +49,55 @@ def to_s
...
+        #
+        # Raised by Encoding and String methods:
+        #   Encoding::UndefinedConversionError:
+        #     when a transcoding operation fails
+        #     if the String contains characters invalid for the target encoding
+        #     e.g. "\x80".encode('UTF-8','ASCII-8BIT')
+        #     and "\x80".encode('UTF-8','ASCII-8BIT', undef: :replace, replace: '<undef>')
+        #     # => '<undef>'
+        #   Encoding::CompatibilityError
+        #    when Enconding.compatbile?(str1, str2) is false
```
bf4: this is technically incorrect.  as written, it's when `nil`.  I'll revise if everything else is ok.

I was confusing [enc.ascii_compatible?](https://github.com/ruby/ruby/blob/ca24e581ba/encoding.c#L493)

```c
* Returns whether ASCII-compatible or not.
 *
 *   Encoding::UTF_8.ascii_compatible?     #=> true
 *   Encoding::UTF_16BE.ascii_compatible?  #=> false
```

with [rb_enc_check](https://github.com/ruby/ruby/blob/ca24e581ba/encoding.c#L844)

```c
rb_enc_check(VALUE str1, VALUE str2)
{
    rb_encoding *enc = rb_enc_compatible(str1, str2);
    if (!enc)
	rb_raise(rb_eEncCompatError, "incompatible character encodings: %s and %s",
```

where [`enc_compatible` is `Encoding.compatible?(str1,str2)](https://github.com/ruby/ruby/blob/ca24e581ba/encoding.c#L1194)

lib/rspec/support/spec.rb
myronmarston: This file is loaded by the `spec_helper` file of each of the rspec repos and we'd like to limit it to stuff that's common setup needed for all the spec suites.  I don't think this module is needed yet in the other repos yet so I'd prefer to not add it to this file yet.  (And hopefully the other repos won't have to deal with encoding stuff directly given that we have `EncodedString` to handle that complexity).

bf4: Ok, will remove once we're done discussing. In the meantime, I'm in CST, it's 12:44am and :zzz:

spec/rspec/support/encoded_string_spec.rb
```diff
@@ -1,51 +1,148 @@
...
+          let(:string) { "\x80".force_encoding(source_encoding) }
+
+          it 'normally raises an Encoding::ConverterNotFoundError' do
+            expect {
+              string.encode(no_converter_encoding)
+            }.to raise_error(Encoding::ConverterNotFoundError)
+          end
+
+          if RUBY_VERSION < '2.1'
+            it 'does nothing' do
```
myronmarston: Why does it do nothing on 1.9 and 2.0?  I didn't think 2.1 had new encoding features...shouldn't it act the same on those rubies?

spec/rspec/support/encoded_string_spec.rb
```diff
@@ -1,51 +1,148 @@
...
+              resulting_string = build_encoded_string(string, no_converter_encoding).to_s
+              expected_string  = EncodedString::REPLACE
+              expect(resulting_string).to be_identical_string(expected_string, no_converter_encoding)
+            end
+          end
+        end
+
+        # see https://github.com/ruby/ruby/blob/34fbf57aaa/transcode.c#L4289
+        # ISO-8859-1 -> UTF-8 -> EUC-JP
+        # "\xa0" NO-BREAK SPACE, which is available in UTF-8 but not in EUC-JP
```
myronmarston: Is `\xa0` available in `ISO-8859-1` encoding?  After all, you're using that encoding below, not UTF-8....

spec/rspec/support/encoded_string_spec.rb
```diff
@@ -1,51 +1,148 @@
...
-            valid_ascii_string = "abc".force_encoding("ASCII-8BIT")
-            valid_unicode_string = utf_8_euro_symbol.force_encoding('UTF-8')
 
-            resulting_string = build_encoded_string(valid_unicode_string, target_encoding) << valid_ascii_string
-            expect(resulting_string).to eq "#{utf_8_euro_symbol}abc".force_encoding('UTF-8')
+            resulting_string = build_encoded_string(valid_unicode_string, utf8_encoding) << valid_ascii_string
+            replacement = EncodedString::REPLACE * 2
+            expected_string = "#{utf_8_euro_symbol}abcd#{replacement}".force_encoding('UTF-8')
+            expect(resulting_string).to be_identical_string(expected_string)
+          end
```
myronmarston: Previously, the spec only demonstrated that it "encodes and appends the string" (as the doc string said), but now it also demonstrates the replacement behavior.  I'd prefer to see the replacement behavior demonstrated in a separate spec that specifically calls out the behavior in a doc string.

Also, it's unclear to me why the `é` gets replaced...that's a valid UTF-8 character, right?

spec/rspec/support/encoded_string_spec.rb
```diff
@@ -1,51 +1,148 @@
...
+            resulting_string = accentless << accented
+            if OS.windows?
+              replacement = "\x82\x82"
+            else
+              replacement = "\u00E9"
+            end
+            expected_string = <<-EOS.encode('UTF-16LE')
+Tu avec carte {count} item has
+Tu avec cart#{replacement} {count} it#{replacement}m has
+            EOS
```
myronmarston: Given that UTF-16 supports accented characters and both strings are in that encoding....why are the accented characters replaced?

spec/rspec/support/encoded_string_spec.rb
```diff
@@ -1,51 +1,148 @@
...
+            expect(resulting_string).to be_identical_string(expected_string)
+          end
+
+          it 'copes with encoded strings' do
+            source_encoding = Encoding.find('UTF-16LE')
+            accentless = build_encoded_string("Tu avec carte {count} item has\n", source_encoding)
+            accented   = "Tu avec carté {count} itém has\n".encode(source_encoding)
+            resulting_string = accentless << accented
+            if OS.windows?
+              replacement = "\x82\x82"
```
myronmarston: What is the windows replacement character and why is it different?  I don't see any windows-specific logic in the implementation...

spec/rspec/support/encoded_string_spec.rb
```diff
@@ -1,51 +1,148 @@
...
+              expect {
+                valid_unicode_string.encode(utf8_encoding) << ascii_string
+              }.to raise_error(Encoding::CompatibilityError)
+            end
+
+            it 'replaces unconvertable characters with a string representation of their hex value' do
+              resulting_string = build_encoded_string(valid_unicode_string, utf8_encoding) << ascii_string
+              expected_string = "#{utf_8_euro_symbol}?"
+              expect(resulting_string).to be_identical_string(expected_string)
+            end
```
myronmarston: This spec confuses me...the doc string says it replaces the characters with their hex value but the example code doesn't seem to demonstrate that...`utf_8_euro_symbol` was defined using hex escape codes but is the actual euro symbol, right?

spec/rspec/support/encoded_string_spec.rb
```diff
@@ -54,23 +151,72 @@ module RSpec::Support
...
+
+        context 'when there is an undefined conversion to the target encoding' do
+          let(:wrapped_string) { "aaaaaaaaaaa#{ascii_arrow_symbol}aaaaa".force_encoding("ASCII-8BIT") }
 
-          expect {
-            build_encoded_string(wrapped_string, target_encoding).split(utf_8_euro_symbol.force_encoding("UTF-8"))
-          }.not_to raise_error
+          it 'normally raises an Encoding::UndefinedConversionError' do
+            expect {
+              wrapped_string.encode(utf8_encoding)
```
myronmarston: This confuses me -- I thought that in this context we were specifying the behavior of `split`, not the behavior of `encode`, so aren't we concerned with what `split` normally does, not what `encode` normally does?

spec/rspec/support/encoded_string_spec.rb
```diff
@@ -54,23 +151,72 @@ module RSpec::Support
...
+            }.to raise_error(Encoding::UndefinedConversionError)
+          end
+
+          it 'splits the string based on the delimiter accounting for encoding' do
+            expect {
+              build_encoded_string(wrapped_string, utf8_encoding).split(utf_8_euro_symbol.force_encoding("UTF-8"))
+            }.not_to raise_error
```
myronmarston: I realize this expectation was already `expect { }.not_to raise error` but it would be nice to instead expect what the return value is as that's a stronger expectation.  (Not a merge blocker since this is how it already was).

spec/rspec/support/encoded_string_spec.rb
```diff
@@ -54,23 +151,72 @@ module RSpec::Support
...
+          end
+        end
+
+        # see https://github.com/rspec/rspec-expectations/blob/f8a1232/spec/rspec/expectations/fail_with_spec.rb#L50
+        #     https://github.com/rspec/rspec-expectations/issues/201
+        #     https://github.com/rspec/rspec-expectations/pull/220
+        context 'with a string that cannot be converted to the target encoding' do
+          let(:binary_poop) {'💩' } # [128169] "\u{1F4A9}"
```
myronmarston: Lame that chrome doesn't render this character :(. (I see it render it my editor fine, though).

myronmarston:
I opened a PR against your branch that refactors the matcher and does a couple other small changes:

bf4/rspec-support#1

lib/rspec/support/spec/encoding_helpers.rb
myronmarston: > Also, if I add a test for this, how/where would I do that?

We definitely test matchers that ship with RSpec but for fairly simple custom matchers like this that we just use in our own specs we don't tend to spec them.  The "test" of them is that we see them pass and fail appropriately.

If you do want to test them, you can see the specs for our built-in matchers as a guide:

https://github.com/rspec/rspec-expectations/tree/master/lib/rspec/matchers/built_in

Basically, we test both positive and negative cases, passing and failing.  You can use `expect { expectation }.to raise_error(RSpec::Matchers::ExpectationNotMetError, /snippet of expected failure/)` (where `expectation` is a code expression that uses the matcher in a way you expect to fail) to test the failing case.

But as I said, I don't consider testing it necessary.

spec/rspec/support/encoded_string_spec.rb
myronmarston: Since this test isn't fixed as part of this PR maybe it should go (actually, probably the whole context) in another PR where you do fix this behavior?

lib/rspec/support/encoded_string.rb
myronmarston: Yeah, please fix this. (Good catch!).  Also, s/compatbile/compatible/

spec/rspec/support/encoded_string_spec.rb
bf4: I picked the `é` up from the differ specs as something that caused passing specs to break when added. Interestingly, RubySpec also uses it https://github.com/rubyspec/rubyspec/blob/91ce9f6549/core/string/fixtures/utf-8-encoding.rb

spec/rspec/support/encoded_string_spec.rb
bf4: I got this from the differ spec.  I'm not sure, but I think it's how converting to UTF-16LE adds empty bytes...

spec/rspec/support/encoded_string_spec.rb
bf4: Just observed.  I looked through the specs wondering if that was a '?' in a CP or IBM encoding but didn't find it.  Not sure.

spec/rspec/support/encoded_string_spec.rb
bf4: Oh, my bad. s/hex/REPLACE. (it was a changed behavior among the churn).  the arrow becomes a '?'

spec/rspec/support/encoded_string_spec.rb
bf4: Sure. I did look at that at one point. When I address your comments tonight, I'll take another look.

bf4:
> I opened a PR against your branch that refactors the matcher and does a couple other small changes:

oh fun. Will review

lib/rspec/support/encoded_string.rb
myronmarston: The detail is helpful, actually.

spec/rspec/support/encoded_string_spec.rb
myronmarston: I'd still like to see that part pulled into a separate spec...and I still don't understand why it gets replaced for a UTF-8 encoded string...

spec/rspec/support/encoded_string_spec.rb
myronmarston: Are you speaking of this differ spec?

https://github.com/rspec/rspec-support/blob/2f8cbe2de08a4f2c248b00b679083334ebadbb0e/spec/rspec/support/differ_spec.rb#L48-L57

It doesn't involve replacements.

spec/rspec/support/encoded_string_spec.rb
myronmarston: Hmm, that's odd, given we pass `?` as the replacement character and are specifying the encoding.  The question is, does this represent a bug or is it correct behavior on windows?  If it's a bug, I don't want to encode that behavior in a spec that makes us think that's how it should work.  I think I'd prefer that we tag this spec with `:failing_on_appveyor` rather than special case windows, and hopefully someone who uses windows as their primary dev environment can take a look later and figure out if it's desirable behavior or not.

spec/rspec/support/encoded_string_spec.rb
myronmarston: Gotcha.  I trust you'll update the doc string then :).

One thing I don't quite understand is the `ascii_arrow_symbol`.  Playing around with it, I see that it isn't valid US-ASCII, and I think of `ASCII-8BIT` as being an alias of `BINARY`...so in what sense is it an arrow symbol?  From what I can tell it's just a byte sequence that's not valid in UTF-8, but maybe I'm missing something...

spec/rspec/support/encoded_string_spec.rb
bf4: I don't know why. Try it in irb. I can't explain it. Maybe it's a bug in 2.0?

```ruby
"\x80".force_encoding("ASCII-8BIT").force_encoding("Emacs-Mule").encode(:invalid => :replace).bytes.to_a
```

spec/rspec/support/encoded_string_spec.rb
bf4: If I understand your question, it's why talk about UTF-8 when we're talking about ISO-8859-1 -> EUC-JP?  If that's your question, [it's because when Ruby determines a conversion path it passes through UTF-8](https://github.com/ruby/ruby/blob/34fbf57aaa/transcode.c#L3119-lL3121)

```ruby
 p Encoding::Converter.search_convpath("ISO-8859-1", "EUC-JP")
   #=> [[#<Encoding:ISO-8859-1>, #<Encoding:UTF-8>],
    #    [#<Encoding:UTF-8>, #<Encoding:EUC-JP>]]
```

and see https://github.com/ruby/ruby/blob/34fbf57aaa/transcode.c#L4242-L4250 and https://github.com/ruby/ruby/blob/34fbf57aaa/transcode.c#L3917-L3973 (and maybe https://github.com/ruby/ruby/blob/34fbf57aaa/transcode.c#L4289-L4294 etc)

And then probably also look at the weirdness that is converter not found: https://github.com/rubyspec/rubyspec/blob/archive/core/string/shared/encode.rb#L96-L101

spec/rspec/support/encoded_string_spec.rb
bf4: per http://stackoverflow.com/questions/21289181/char174-returning-the-value-of-char0174-why which links to http://www.theasciicode.com.ar/extended-ascii-code/angle-quotes-guillemets-right-pointing-double-angle-french-quotation-marks-ascii-code-174.html

> It is all about character encoding. 174 (AE in hex) is ® in Unicode, but it is « in Extended ASCII code.

e.g. CP1252 or Windows-1251 or something like that
But I didn't add that specific string and haven't looked into the origins in the suite.

bf4:
Didn't finish addressing comments tonight.  Merged and amended last commit in your PR.  Will finish and push code 'tomorrow' :zzz:

spec/rspec/support/encoded_string_spec.rb
myronmarston: Wow, I had no idea just how weird and complex ruby encoding behavior is!

spec/rspec/support/encoded_string_spec.rb
myronmarston: Huh, weird.  You mind leaving a comment that provides this snippet to demonstrate the different 2.0 vs 2.1+ behavior?

bf4:
```bash
LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 \
 rvm ruby-1.9.2-p330,ruby-1.9.3-p551,ruby-2.0.0-p598,ruby-2.1.5,ruby-2.2.0,jruby-1.7.18,rbx-2.2.2 do \
 ruby -e 'p [RUBY_VERSION, RUBY_ENGINE, Encoding.default_external, __ENCODING__] + "\x80".force_encoding("ASCII-8BIT").force_encoding("Emacs-Mule").encode(:invalid => :replace).bytes.to_a'

 for version in 1.9 2.0; do \
 export JRUBY_OPTS="-Xcompat.version=${version}" ; \
 LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 \
  rvm jruby-1.7.18 do \
 ruby -e 'p [RUBY_VERSION, RUBY_ENGINE, Encoding.default_external, __ENCODING__] + "\x80".force_encoding("ASCII-8BIT").force_encoding("Emacs-Mule").encode(:invalid => :replace).bytes.to_a' ; done
 ```

```plain
["1.9.2", "ruby", #<Encoding:UTF-8>, #<Encoding:UTF-8>, 128]
["1.9.3", "ruby", #<Encoding:UTF-8>, #<Encoding:UTF-8>, 128]
["2.0.0", "ruby", #<Encoding:UTF-8>, #<Encoding:UTF-8>, 128]
["2.1.5", "ruby", #<Encoding:UTF-8>, #<Encoding:UTF-8>, 63]
["2.2.0", "ruby", #<Encoding:UTF-8>, #<Encoding:UTF-8>, 63]
["2.0.0", "jruby", #<Encoding:UTF-8>, #<Encoding:UTF-8>, 128]
["2.1.0", "rbx", #<Encoding:UTF-8>, #<Encoding:UTF-8>, 128]

["1.9.3", "jruby", #<Encoding:UTF-8>, #<Encoding:UTF-8>, 128]
["2.0.0", "jruby", #<Encoding:UTF-8>, #<Encoding:UTF-8>, 128]
```

bf4: fwiw I just asked about this on the ruby forum https://www.ruby-forum.com/topic/6861247

bf4:
Didn't work on it today.  Probably won't again till Sunday.

myronmarston:
That's fine :).

spec/rspec/support/encoded_string_spec.rb
bf4: Figured it out, or at least the meaning of `\x82`:

```ruby
egrave =  "é".encode('UTF-8') #=> "é"
egrave.bytes #=>  [195, 169]
encoded_egrave = egrave.encode('utf-16le') #=> "\u00E9" 
encoded_egrave.bytes #=> > [233, 0] 
# but "\x82" aka 130 is egrave in extended ascii, so on windows
"\x82".force_encoding('ibm437').encode('utf-8') #=> 'é'
# and thus we can reverse it to 
egrave.encode('ibm437') #=> "\x82"
```

Though I'm not sure why it would be encoding it as `ibm437` on windows, and not `utf-16le`.

bf4: Hmm, looking at http://stackoverflow.com/questions/1259084/what-encoding-code-page-is-cmd-exe-using and https://github.com/ruby/ruby/blob/9fd7afefd04134c98abe594154a527c6cfe2123b/ext/win32ole/win32ole.c#L540, it appears you can get the current encoding (codepage) on windows in the command prompt by running `chcp` and change it to utf8 via `chcp 65001`.

I'd like to add that and `export LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8` as documented pre-conditions for running the test suite.

bf4:
Rebased off of current master and force pushed

Add new commits to PR per discussion for easy review, if ok, would like to discuss which concepts in the pr deserve their own commit and how your PR into my PR should appear. (if you don't care, I'm comfortable just doing something).

Also, I looked into windows encodings more and think I'd like to document how to set the default external encoding that the suite expects on posix and windows, see https://github.com/rspec/rspec-support/pull/151#discussion_r22637355 and https://github.com/rspec/rspec-support/pull/151#discussion_r22991439.

bf4:
@myronmarston This is ready for review, when you get the chance.  (Also see previous comment).

myronmarston:
Merged.  Thanks, @bf4!

bf4:
:rainbow: :fireworks: :horse: :100: yay!

lib/rspec/support/encoded_string.rb
```diff
@@ -33,17 +47,55 @@ def to_s
...
+        #
+        # Raised by Encoding and String methods:
+        #   Encoding::UndefinedConversionError:
+        #     when a transcoding operation fails
+        #     if the String contains characters invalid for the target encoding
+        #     e.g. "\x80".encode('UTF-8','ASCII-8BIT')
+        #     vs "\x80".encode('UTF-8','ASCII-8BIT', undef: :replace, replace: '<undef>')
+        #     # => '<undef>'
+        #   Encoding::CompatibilityError
+        #    when Enconding.compatbile?(str1, str2) is false
```
sj26: Maybe should be Encoding. :-)

bf4: Prol'y

spec/rspec/support/encoded_string_spec.rb
sj26: http://thenextweb.com/google/2015/01/11/emoji-support-finally-coming-google-chrome-os-x/

lib/rspec/support/encoded_string.rb
sj26: Nice docs. There's not enough about ruby encodings around the place so this is great.

PR #157, opened 2015-01-16
Don’t consider empty arrays and empty hashes to match.

### LINKED ISSUE #156: Hash/Array include matcher inconsistency (fuzzy matcher)
I'm filing this here and not in rspec-expectations because the bug is in [fuzzy_matcher.rb#L9](https://github.com/rspec/rspec-support/blob/master/lib/rspec/support/fuzzy_matcher.rb#L9).

```ruby
RSpec.describe "Example" do
  it "should fail" do
    # fails correctly:
    expect("data" => []).to include("data" => {})

    # passes incorrectly:
    expect("data" => {}).to include("data" => [])
  end
end
```

In the expectation that incorrectly passes, the code falls through the `if` branch on line 9 and compares the two objects as if they were both arrays, although one is a hash and the other is an array. It proceeds into `arrays_match?` and since they are both empty it passes.

myronmarston:
Fixes #156.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,11 @@ Enhancements:
 
 * Add extra Ruby type detection. (Jon Rowe, #133)
 
+Bug Fixes:
+
+* Do not consider `[]` and `{}` to match when performing fuzzy matching.
+  (Myron Marston, #157)
+
 ### 3.1.2 / 2014-10-08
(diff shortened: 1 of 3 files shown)


myronmarston:
@rspec/rspec -- can someone review this?

JonRowe:
LGTM :)

PR #158, opened 2015-01-17
Include details on how to run against `master`.

cupakromer:
A very common issue is when a user attempts to run against the `master`
branch. Simply pointing to `master` doesn't work. This adds the steps
necessary to use `master`.

We need all of the repos in order to also satisfy any other gems which
may have a dependency on the `rspec` gem.

[ci skip]

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,15 @@ suitable for internal use only at this time.
 
 Install one or more of the `RSpec` gems.
 
+Want to run against the `master` branch? You'll need to include the dependent
+RSpec repos as well. Add the following to your `Gemfile`:
+
+```ruby
+%w[rspec-core rspec-expectations rspec-mocks rspec-support].each do |lib|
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,6 +8,15 @@ suitable for internal use only at this time.
...
 
+Want to run against the `master` branch? You'll need to include the dependent
+RSpec repos as well. Add the following to your `Gemfile`:
+
+```ruby
+%w[rspec-core rspec-expectations rspec-mocks rspec-support].each do |lib|
+  gem lib, :git => "git://github.com/rspec/#{lib}.git", :branch => 'master'
+end
+```
+
```
myronmarston: I think it's less likely that people are gonna be aware of needing to do this for rspec-support since it doesn't have any public APIs.  I'm fine including this, though.

myronmarston:
LGTM

README.md
cupakromer: I think it's good parity and for the few cases where we ask someone to try a fix here.

PR #160, opened 2015-01-29
Allow differ instance to be used multiple times

waterlink:
related to this conversation: https://github.com/rspec/rspec-expectations/pull/713#discussion_r23645315

Story short: current implementation of `Differ` relies on `@hunks` to output diffs, it uses method `hunks` that has memoizing inside `@hunks ||= ...`, which means when `Differ` instance used subsequently it will use `@hunks` from the first run. Since `Differ` has only one method you usually call it with (ie `#diff`) I have put an `@hunks = nil` there, so it will instantiate new `@hunks` each time.

I have considered creating additional class which will be instantiated in each run of `#diff`and basically all calculation will be delegated to it. This could make this a bit clearer, but on the other hand a bit slower. WDYT?

/c @rspec/rspec @myronmarston

### DIFF
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -25,12 +25,13 @@ module RSpec
 
       # rubocop:disable MethodLength
       def diff_as_string(actual, expected)
-        @encoding = pick_encoding actual, expected
+        encoding = pick_encoding(actual, expected)
 
-        @actual   = EncodedString.new(actual, @encoding)
-        @expected = EncodedString.new(expected, @encoding)
+        actual   = EncodedString.new(actual, encoding)
+        expected = EncodedString.new(expected, encoding)
 
-        output = EncodedString.new("\n", @encoding)
+        output = EncodedString.new("\n", encoding)
+        hunks = build_hunks(actual, expected)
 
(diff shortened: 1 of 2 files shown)


myronmarston:
Thanks for looking into this!

I'm not really concerned about the speed of having another class that you delegate to, particularly since diffs are only rendered in a failing spec which is generally less common than a passing spec, so I think I'd be in favor doing the extraction you've suggested, with one caveat: having never worked in this area, I'm not sure how good the test coverage is and doing a refactoring with coverage that potentially has holes carries some risk.  One potential middle ground (which I'd favor over what you've done here) is a more modest refactoring: keep things all in the same class, but remove all instance variables all together, opting to pass arguments to other methods rather than using instance variables as a way to pass state to a later method call without actually passing it as an argument.

waterlink:
@myronmarston thx, I will go for that one then.

waterlink:
@myronmarston I have updated the PR with your suggestion

myronmarston:
Great work, thanks! That's very easy to reason about :).

myronmarston:
@waterlink, care to take a stab at updating rspec-expectations so that in the case we had discussed, it re-uses the differ?

waterlink:
@myronmarston yes, I was going to do it.

Issue #163, opened 2015-01-31
Spec ./spec/rspec/support/encoded_string_spec.rb:98 passes on MRI when it shouldn't


YorickPeterse:
This is an odd case. The spec https://github.com/rspec/rspec-support/blob/master/spec/rspec/support/encoded_string_spec.rb#L98 seems to pass on MRI and fail on Rubinius, but it makes little sense as to why it passes on MRI.

The problem on Rubinius seems to be that it's comparing `"\x80"` with `"?"`, both encoded using Emacs-Mule. So see what data is used I "instrumented" the code using good ol' `p`:

```diff
diff --git a/spec/rspec/support/encoded_string_spec.rb b/spec/rspec/support/encoded_string_spec.rb
index 6c2b6ce..53c242c 100644
--- a/spec/rspec/support/encoded_string_spec.rb
+++ b/spec/rspec/support/encoded_string_spec.rb
@@ -98,6 +98,11 @@ module RSpec::Support
             it 'forces the encoding and replaces invalid characters with the REPLACE string' do
               resulting_string = build_encoded_string(string, no_converter_encoding).to_s
               expected_string  = EncodedString::REPLACE.force_encoding(no_converter_encoding)
+
+              p build_encoded_string(string, no_converter_encoding).to_s
+              p resulting_string
+              p expected_string
+
               expect(resulting_string).to be_identical_string(expected_string)
             end
           end
```

On MRI the output of this test is as following:

```
Run options: include {:focus=>true, :locations=>{"./spec/rspec/support/encoded_string_spec.rb"=>[98]}}

Randomized with seed 6728

RSpec::Support::EncodedString
  #to_s
    when no converter is known for an encoding
"\x80"
"?"
"?"
      forces the encoding and replaces invalid characters with the REPLACE string

Finished in 0.0013 seconds (files took 0.08671 seconds to load)
1 example, 0 failures

Randomized with seed 6728
```

On Rubinius the output is as following:

```
Run options: include {:focus=>true, :locations=>{"./spec/rspec/support/encoded_string_spec.rb"=>[98]}}

Randomized with seed 40770

RSpec::Support::EncodedString
  #to_s
    when no converter is known for an encoding
"\x80"
"\x80"
"?"
      forces the encoding and replaces invalid characters with the REPLACE string (FAILED - 1)

Failures:

  1) RSpec::Support::EncodedString#to_s when no converter is known for an encoding forces the encoding and replaces invalid characters with the REPLACE string
[...]

PR #164, opened 2015-01-31
Don't modify EncodedString::REPLACE in place.

YorickPeterse:
I discovered this when looking into #163, though it doesn't fix the problem described in said issue just yet.


myronmarston:
Good catch...I forgot `force_encoding` modifies the string in place.  It's too bad there's not `force_encoding` and `force_encoding!` to force you to explicitly opt-in to mutation :(.

PR #165, opened 2015-01-31
Dont mutate input string

myronmarston:
@YorickPeterse I think this fixes the case you noted where we were mutating the input string via `force_encoding`.


JonRowe:
LGTM

PR #166, opened 2015-02-01
Update the default `skip_frames` value.

myronmarston:
There are 2 more frames we can safely skip. Skipping
these by default increases the chance we’ll hit the first
non-rspec line in the first group of frames, which will
perform better.


JonRowe:
LGTM

PR #167, opened 2015-02-03
Test Differ#pick_encoding; move to EncodedString

bf4:
Followup to https://github.com/rspec/rspec-support/pull/151

especially see https://github.com/rspec/rspec-support/pull/134#issuecomment-68984440

### DIFF
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -25,7 +25,7 @@ module RSpec
 
       # rubocop:disable MethodLength
       def diff_as_string(actual, expected)
-        encoding = pick_encoding(actual, expected)
+        encoding = EncodedString.pick_encoding(actual, expected)
 
         actual   = EncodedString.new(actual, encoding)
         expected = EncodedString.new(expected, encoding)
@@ -187,15 +187,6 @@ module RSpec
           object =~ /\n/ ? object : object.inspect
         else
           PP.pp(object, "")
-        end
-      end
-
-      if String.method_defined?(:encoding)
-        def pick_encoding(source_a, source_b)
-          Encoding.compatible?(source_a, source_b) || Encoding.default_external
-        end
-      else
-        def pick_encoding(_source_a, _source_b)
         end
       end
 
--- a/lib/rspec/support/encoded_string.rb
+++ b/lib/rspec/support/encoded_string.rb
@@ -105,7 +105,14 @@ module RSpec
         def detect_source_encoding(string)
           string.encoding
         end
+
+        def self.pick_encoding(source_a, source_b)
+          Encoding.compatible?(source_a, source_b) || Encoding.default_external
+        end
       else
+
+        def self.pick_encoding(_source_a, _source_b)
+        end
 
         private
 
--- a/spec/rspec/support/differ_spec.rb
+++ b/spec/rspec/support/differ_spec.rb
@@ -7,6 +7,33 @@ require 'rspec/support/differ'
 module RSpec
   module Support
     describe Differ do
+      describe '#pick_encoding' do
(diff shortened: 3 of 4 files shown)


lib/rspec/support/differ.rb
```diff
@@ -7,6 +7,7 @@ module RSpec
   module Support
     # rubocop:disable ClassLength
     class Differ
+      LINE_BREAK = "\n"
```
JonRowe: What's the reasoning behind this?

bf4: Well, as I commented in https://github.com/bf4/rspec-support/commit/ce333af71a110fe5f080b3b0e8cc37c8cb05db6a I was thinking of using `LINE_BREAK` as a way to get the `__ENCODING__` of strings generated in `differ.rb`, that also happened to be a nice cleanup of string generation. Turns out it didn't work like I thought it did, so, the only reason to keep it now is cleanup, if you like. I'm going to try again with a less aggressive strategy.

lib/rspec/support/differ.rb
```diff
@@ -95,18 +96,18 @@ def diffably_stringify(array)
           if Array === entry
             entry.inspect
           else
-            entry.to_s.gsub("\n", "\\n")
+            entry.to_s.gsub(LINE_BREAK, Regexp.escape(LINE_BREAK))
```
JonRowe: So the pre-escaped `"\\n"` still seems like a better idea than running an escape each time this is used.

lib/rspec/support/differ.rb
```diff
@@ -182,9 +183,9 @@ def object_to_string(object)
             pp_value = PP.singleline_pp(object[key], "")
 
             "#{pp_key} => #{pp_value},"
-          end.join("\n")
+          end.join(LINE_BREAK)
         when String
-          object =~ /\n/ ? object : object.inspect
+          object =~ /#{LINE_BREAK}/o ? object : object.inspect
```
JonRowe: Same here, leaving the primitive seems better than interpolating the constant each time.

bf4:
I thinking of simplifying this PR to 

1. Testing `Differ#pick_encoding`
2. Moving `pick_encoding` into `EncodedString`

Then, in another PR, per any discussion in this one, change the tests in the `differ_spec.rb` to use an encoding-agnostic `be_identical_string` matcher, such as in this [changset](https://github.com/bf4/rspec-support/commit/6ab674a9fbc9102daf1aaa411413fb1ffa4d46e8#diff-298595d31b0fe8bb190912673bab6659R28-R35), or just use object equality, for now.

bf4:
Since I'm going to revise this PR a bit, I thought I might as well push the ArgumentError on EncodingString#split fix for a preview of what it is, though it would go into a subsequent PR.

lib/rspec/support/spec/encoded_string_matcher.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  chain :with_encoding do |enc|
+    @expected_encoding = enc
+  end
+
+  # HACK: To disable the encoding check
+  chain :without_encoding do
+    @without_encoding = :without_encoding
+  end
```
bf4: re: the hack here and below, I'm wondering if I should change the name of the matcher to `expect_identical_encoded_string` and `expect_identical_string` and see if there are still any cases for chaining

bf4:
I want to keep this moving, so I've reduced this PR to just the Differ/EncodedString#pick_encoding

JonRowe:
Thanks @bf4

JonRowe:
(Note @myronmarston I haven't added a change log because this is pretty much just an internal refactoring)

PR #173, opened 2015-02-08
The Differ no longer catches CompatibilityErrors

bf4:
This code was introduced in 
https://github.com/rspec/rspec-expectations/pull/220/files#diff-9533f5f156a38a3307ecfc610d2282d7R47

but is no longer raised either by the the current test code, the original test
code
 @expected="Tu avec carté {count} itém has".encode('UTF-16LE')
 @actual="Tu avec carte {count} item has".encode('UTF-16LE') or any other
variations I've tried.

So, I'm proposing to remove it, especially since encoding
is well-supported and tested by EncodedString, which the Differ relies on.

### DIFF
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -48,8 +48,6 @@ module RSpec
         finalize_output(output, hunks.last.diff(format_type).to_s) if hunks.last
 
         color_diff output
-      rescue Encoding::CompatibilityError
-        handle_encoding_errors(actual, expected)
       end
       # rubocop:enable MethodLength
 
@@ -187,17 +185,6 @@ module RSpec
           object =~ /\n/ ? object : object.inspect
         else
           PP.pp(object, "")
-        end
-      end
-
-      def handle_encoding_errors(actual, expected)
-        if actual.source_encoding != expected.source_encoding
-          "Could not produce a diff because the encoding of the actual string " \
-          "(#{actual.source_encoding}) differs from the encoding of the expected " \
-          "string (#{expected.source_encoding})"
-        else
-          "Could not produce a diff because of the encoding of the string " \
-          "(#{expected.source_encoding})"
         end
       end
     end
--- a/spec/rspec/support/differ_spec.rb
+++ b/spec/rspec/support/differ_spec.rb
@@ -112,17 +112,6 @@ EOD
             expect(differ.diff(actual, expected)).to eq("\n@@ -1,2 +1,2 @@\n-Tu avec carte {count} item has\n+Tu avec carté {count} itém has\n")
             expect(differ.diff(actual, expected).encoding).to eq(Encoding.default_external)
           end
-
-          it 'handles any encoding error that occurs with a helpful error message' do
-            expect(RSpec::Support::HunkGenerator).to receive(:new).
-              and_raise(Encoding::CompatibilityError)
-            expected = "Tu avec carte {count} item has\n".encode('us-ascii')
-            actual   = "Tu avec carté {count} itém has\n"
-            diff = differ.diff(actual, expected)
-            expect(diff).to match(/Could not produce a diff/)
-            expect(diff).to match(/actual string \(UTF-8\)/)
-            expect(diff).to match(/expected string \(US-ASCII\)/)
-          end
         end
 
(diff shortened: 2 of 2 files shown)


JonRowe:
Sorry I don't want to remove this, I have seen this triggered in issues people have opened and its notoriously hard to replicate on a standard system so I'd rather have this safety net.

bf4:
Np, just wanted to put it out there for reference.  I'd love to test if you see examples.

JonRowe:
If you'd like to fake up a spec for it to show what it does when triggered I'd merge, but I've had difficulty triggering this on Travis without stubbing out a call.

bf4:
I spent quite a bit of time trying to get the differ to raise this exception and never succeeded, so, maybe we should change the message to

>  If you are reading this error message, please report it with backtrace and string to rspec-support 173  :)

myronmarston:
> Sorry I don't want to remove this, I have seen this triggered in issues people have opened and its notoriously hard to replicate on a standard system so I'd rather have this safety net.

I lean the other way -- I really hate leaving code in place "just in case".  If we cannot find a way to replicate that particular exception I'd rather not continue to have the baggage of the (apparently) unneeded `rescue`.

I think the worst case to removing it is some user gets an error, and they report it, which allows us to turn that into an actual regression spec, fix it, and get a patch release out quickly with the fix.

If we do decide to keep that `rescue`, I think it really belongs in `EncodedString`, not in the differ...after all, the entire point of `EncodedString` is to deal with these encodings issues so that other places in RSpec don't have to do rescues like this.  Dealing with encodings is entirely outside the responsibility of the differ.

JonRowe:
> I lean the other way -- I really hate leaving code in place "just in case". If we cannot find a way to replicate that particular exception I'd rather not continue to have the baggage of the (apparently) unneeded rescue.

Normally I'd agree with you, but this is an observed edge case, just one we can't replicate on a standard UTF-8 machine, but saying we should remove it because we can't replicate it is like saying we should remove our windows fixes because we can't replicate them on OS X.

> If we do decide to keep that rescue, I think it really belongs in EncodedString, not in the differ...after all, the entire point of EncodedString is to deal with these encodings issues so that other places in RSpec don't have to do rescues like this. Dealing with encodings is entirely outside the responsibility of the differ.

I'm ok with moving it...

myronmarston:
> Normally I'd agree with you, but this is an observed edge case, just one we can't replicate on a standard UTF-8 machine, but saying we should remove it because we can't replicate it is like saying we should remove our windows fixes because we can't replicate them on OS X.

Well, if you're sure that the error can still occur that's different.  I thought we were at the point where we thought it couldn't but you wanted to leave the rescue in place "just in case".  As for the standard UTF-8 bit...I believe we can write specs (potentially that shell out to a new process if necessary) that have different default encodings in order to replicate such things.

> I'm ok with moving it...

I'd like to see that.  As I said, it feels like an SRP violation and would also likely require us to rescue that error elsewhere if/when we start using `EncodedString` elsewhere.

bf4:
We'd likely need to introduce more `EncodedString` objects into the `differ.rb`.  The compatibility error, is, [I think tested against in EncodedString](https://github.com/rspec/rspec-support/blob/ff3c0a609bf959f82787e9b9256810f8efe6c8f0/spec/rspec/support/encoded_string_spec.rb#L177-L188).  Also, we can totally avoid it with [`Encoding.compatible?(a,b)`.](https://github.com/rspec/rspec-support/blob/ff3c0a609bf959f82787e9b9256810f8efe6c8f0/lib/rspec/support/encoded_string.rb#L72), via [#pick_encoding](https://github.com/rspec/rspec-support/blob/ff3c0a609bf959f82787e9b9256810f8efe6c8f0/lib/rspec/support/encoded_string.rb#L135)

bf4:
My 2 :moneybag:  is that it's a judgement call which way to go, and either way should to lead to a user report if there's a failure.

JonRowe:
My concern is that `diff-lcs` can (and will) generate this error if we provide it with encoding that isn't compatible with what it uses (and it will also have internally encoded strings based on any magic comments they use) which is what I think has triggered this in the past (we've encoded the strings accordingly to the user but thats conflicted with `diff-lcs`.

PR #174, opened 2015-02-08
Differ tests no longer use Differ to report diff expectation

bf4:
For example, previously we would write our differ_spec expectations as

```ruby
 actual = differ.diff(str1, str2)
 expect(actual).to eq(expected)
```

 Which had the suprising (but expected) result
of failure diffing changing the string

```ruby
 actual = differ.diff(str1, str2)
 actual != expected &&
   RSpec::Expectations.fail_with(message, actual, expected)
```

`RSpec::Expectations.fail_with` calls `differ.diff(actual, expected)`
 where differ is 

```plain
**`RSpec::Support::Differ.new**(
   :object_preparer => lambda { |object|
RSpec::Matchers::Composable.surface_descriptions_in(object) },
   :color => RSpec::Matchers.configuration.color?
  )
```

So, all these differ_spec.rb tests where we say `expect(str1).to eq(str2)` will
run through the differ twice on failure.

Rather than test them as `expect(str1 == str2).to eq(true)`, use the `#be_correctly_diffed` matcher.

We're not testing encoding because strings created in differ_spec.rb have a
default encoding of UTF-8 in all cases, but string created in the differ.rb
may be US-ASCII in MRI prior to 2.0 else UTF-8. 
https://github.com/ruby/ruby/blob/aacc35e144/encoding.c#L1741

Refs:
- https://github.com/rspec/rspec-support/pull/167
- https://github.com/rspec/rspec-support/pull/151
- https://github.com/rspec/rspec-support/pull/134#issuecomment-68984440

### DIFF
--- a/spec/rspec/support/differ_spec.rb
+++ b/spec/rspec/support/differ_spec.rb
@@ -4,6 +4,32 @@ require 'ostruct'
 require 'timeout'
 require 'rspec/support/differ'
 
+# Special matcher for comparing strings so that
+# we don't run any expectation failures through the Differ.
+# Instead, confirm the strings have the same bytes.
+RSpec::Matchers.define :be_diffed_as do |expected|
+
+  if String.method_defined?(:encoding)
+    match do
+      actual.bytes.to_a == expected.bytes.to_a
+    end
+
+    failure_message do
+      "expected\n#{actual.inspect} to be identical to\n#{expected.inspect}\n"\
+        "The exact bytes are printed below for more detail:\n"\
+        "#{actual.bytes.to_a}\n"\
+        "#{expected.bytes.to_a}\n"\
+    end
+  else
+    match do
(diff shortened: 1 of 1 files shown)


JonRowe:
Ok I like this approach so this isn't a block, but I'm wondering if "correctly" diffed is the right name for the matcher, `be_diffed_as(expected)` seems more natural.

bf4:
@JonRowe renamed.  There is duplication with the matcher in encoded_string_spec but I figure that can be a later refactor if desired.

JonRowe:
Thanks, would you like to tackle the duplication?

bf4:
Sure. What do you think about putting it in `lib/rspec/support/spec/string_matcher.rb` and requiring it directly in the two specs?jkl;

JonRowe:
Sure

bf4:
ref #176

PR #176, opened 2015-02-09
Unify be_identical_string/be_diffed_as in spec/string_matcher.rb

bf4:
Changed meaning of be_identical_string not to include an encoding comparison. 
Encoding comparison is now only turned on by chaining
'with_same_encoding'.

This leaves open the possibility of chaining 'with_encoding(enc)', but there's
currently not need for that.

### DIFF
--- a/lib/rspec/support/spec/string_matcher.rb
+++ b/lib/rspec/support/spec/string_matcher.rb
@@ -1,0 +1,46 @@
+require 'rspec/matchers'
+# Special matcher for comparing encoded strings so that
+# we don't run any expectation failures through the Differ,
+# which also relies on EncodedString. Instead, confirm the
+# strings have the same bytes.
+RSpec::Matchers.define :be_identical_string do |expected|
+
+  if String.method_defined?(:encoding)
+    match do
+      expected_encoding? &&
+        actual.bytes.to_a == expected.bytes.to_a
+    end
+
+    failure_message do
+      "expected\n#{actual.inspect} (#{actual.encoding.name}) to be identical to\n"\
+        "#{expected.inspect} (#{expected.encoding.name})\n"\
+        "The exact bytes are printed below for more detail:\n"\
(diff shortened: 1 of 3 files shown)


lib/rspec/support/spec/string_matcher.rb
```diff
@@ -0,0 +1,45 @@
...
+        "The exact bytes are printed below for more detail:\n"\
+        "#{actual.bytes.to_a}\n"\
+        "#{expected.bytes.to_a}\n"\
+    end
+
+    # Depends on chaining :with_same_encoding for it to
+    # check for string encoding.
+    def expected_encoding?
+      if defined?(@expect_same_encoding) && @expect_same_encoding
+        actual.encoding == expected.encoding
```
JonRowe: Won't this blow up on 1.8.7?

JonRowe: Nevermind

bf4:
forgot to require "rspec/matchers" at the top of the file. amended and force pushed.

lib/rspec/support/spec/string_matcher.rb
bf4: I'm really not sure how to get the best semantics from the matchers and also the least hacky implementation of the matcher.  I'm not married to this implementation, but I liked the semantics of `be_identical_string(str).with_same_encoding` better than adding a new `be_identical_encoded_string` that is called from `be_identical_string`.

myronmarston:
LGTM

lib/rspec/support/spec/string_matcher.rb
JonRowe: I just missed the encoding check above

JonRowe:
Thanks @bf4 :)

bf4:
Wow, that was quite a weekend. Thanks @JonRowe @myronmarston

PR #178, opened 2015-02-10
Add `in_sub_process_where_possible`

JonRowe:
For specs where we care about not poisoning our environment it'd be good to run
them in a subprocess, but these specs may be too important to skip on
platforms without fork, so we can just run normally instead.

### DIFF
--- a/lib/rspec/support/spec/in_sub_process.rb
+++ b/lib/rspec/support/spec/in_sub_process.rb
@@ -35,13 +35,18 @@ module RSpec
 
           raise exception if exception
         end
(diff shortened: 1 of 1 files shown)


myronmarston:
I think I'd prefer this was called `in_sub_process_if_possible`...it's more of a conditional thing than a location thing and "if" is shorter, which is nice :).

LGTM otherwise.

JonRowe:
Changed, always up for shorter ;)

myronmarston:
LGTM

PR #179, opened 2015-02-11
WIP: Update shared context to minimize loaded stdlibs.

### LINKED ISSUE #177: Add shared example group that we can use in each repo to enforce minimization of stdlib usage
(Continuing from rspec/rspec-expectations#730...)

Given our goal of minimizing how much of the stdlib we use (so as to not to give users a green suite when they forget to load the bits they use), it would be useful to enforce it via some specs that enforce an "allowed stdlibs" whitelist.  Using something new from stdlib would require an intentional decision to add to that list and hopefully over time we can shrink it.  It could also be used to help us accurately document what stdlib pieces we do use.

My idea for how to do this is similar to the warning prevention.  We can use `$"` (aliased as `$LOADED_FEATURES` when `english` is loaded) to get a list of what all has been loaded:

```
➜  ~  ruby -e 'puts $"' --disable=gem
enumerator.so
/Users/myron/.rubies/ruby-2.1.5/lib/ruby/2.1.0/x86_64-darwin12.0/enc/encdb.bundle
/Users/myron/.rubies/ruby-2.1.5/lib/ruby/2.1.0/x86_64-darwin12.0/enc/trans/transdb.bundle
```

So, we can run that and capture what is loaded by default, and then run that again with all the lib files from a particular rspec project loaded.  We can subtract them to get everything loaded by that project.  We can further subtract off all the RSpec internal files. That should leave all the external stuff (stdlibs and external gems like diff-lcs) that are currently being loaded, and we can easily compare that to a whitelist and fail if they get out of sync.

myronmarston:
Fixes #177.

More to come here -- I have some other ideas -- but I ran out of time this morning and wanted to push what I had to get travis builds going.

/cc @JonRowe

### DIFF
--- a/lib/rspec/support/ruby_features.rb
+++ b/lib/rspec/support/ruby_features.rb
@@ -1,3 +1,5 @@
+require 'rbconfig'
+
 module RSpec
   module Support
     # @api private
--- a/lib/rspec/support/spec/library_wide_checks.rb
+++ b/lib/rspec/support/spec/library_wide_checks.rb
@@ -5,6 +5,7 @@ RSpec.shared_examples_for "library wide checks" do |lib, options|
   allowed_loaded_feature_regexps = options.fetch(:allowed_loaded_feature_regexps, [])
   preamble_for_lib = options[:preamble_for_lib]
   preamble_for_spec = "require 'rspec/core'; require 'spec_helper'"
+  skip_spec_files = options.fetch(:skip_spec_files, /MATCHES NOTHING/)
 
   include RSpec::Support::ShellOut
 
@@ -55,6 +56,7 @@ RSpec.shared_examples_for "library wide checks" do |lib, options|
 
   define_method :load_all_spec_files do
     files = files_to_require_for("spec") + lib_test_env_files
+    files = files.reject { |f| f =~ skip_spec_files }
     load_all_files(files, preamble_for_spec)
   end
 
--- a/lib/rspec/support/spec/library_wide_checks.rb
+++ b/lib/rspec/support/spec/library_wide_checks.rb
@@ -1,0 +1,120 @@
+require 'rspec/support/spec/shell_out'
+
+RSpec.shared_examples_for "library wide checks" do |lib, options|
+  consider_a_test_env_file = options.fetch(:consider_a_test_env_file, /MATCHES NOTHING/)
+  allowed_loaded_feature_regexps = options.fetch(:allowed_loaded_feature_regexps, [])
+  preamble_for_lib = options[:preamble_for_lib]
+  preamble_for_spec = "require 'rspec/core'; require 'spec_helper'"
+
+  include RSpec::Support::ShellOut
+
+  define_method :files_to_require_for do |sub_dir|
+    slash         = File::SEPARATOR
+    lib_path_re   = /#{slash + lib}[^#{slash}]*#{slash}lib/
+    load_path     = $LOAD_PATH.grep(lib_path_re).first
+    directory     = load_path.sub(/lib$/, sub_dir)
+    files         = Dir["#{directory}/**/*.rb"]
+    extract_regex = /#{Regexp.escape(directory) + File::SEPARATOR}(.+)\.rb$/
+
+    # We sort to ensure the files are loaded in a consistent order, regardless
+    # of OS. Otherwise, it could load in a different order on Travis than
+    # locally, and potentially trigger a "circular require considered harmful"
+    # warning or similar.
+    files.sort.map { |file| file[extract_regex, 1] }
+  end
+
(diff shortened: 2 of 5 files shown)


JonRowe:
Seems... complex? Not sure I grok it atm

myronmarston:
> Seems... complex? Not sure I grok it atm

Which parts are you confused about?  I adapted most of this from https://github.com/rspec/rspec-support/blob/master/lib/rspec/support/spec/prevent_load_time_warnings.rb which we'll be removing once core/expectations/mocks have been updated to use this new example group.  Notable changes from that are:

* Besides tracking warnings when all `lib` files of the gem are loaded, we also track `$"` (aka `$LOADED_FEATURES`) which keeps track of everything that's been loaded.  We can then use this to enforce a white list of allowed things that a particular project loads.  Over time hopefully we can shrink this list and it ensures we think consciously about it before adding a new stdlib dependency.
* The specs from the other shared example group are so much slower than the typical core/expectations/mocks specs, so I did a few things here to speed things up:
  * When we shell out to load all lib files we have it track both warnings and loaded features as part of the same spawned ruby program -- that way we don't have to load all lib files twice.  We do it in a `before(:context)` hook so that multiple examples can inspect the results.
  * Since spawning the "load all lib files" and "load spec files" ruby programs is all external I/O, I decided to parallelize them with threads.  This speeds things up.
* Now that this shared example group isn't just about preventing warnings, but is more generally about library-wide checks, I decided to add a thing that prevents malformed whitespace as well.  I've seen some PRs from time to time where users use tabs instead of spaces or leave trailing whitespace.

Does that explain it better?

JonRowe:
Ah ok, I didn't realise Ruby tracked requires in that way... TIL.. :D

myronmarston:
So the appveyor build is failing and I'm not sure how to fix it.  The new "loads only a known set of stdlibs" spec fails with:

>      Failure/Error: loaded_features = File.read(loaded_features_outfile).split("\n")
>     Errno::ENOENT:
>       No such file or directory - ./loaded_features.txt

The "issues no warnings when loaded" spec fails with:

>   Failure/Error: expect(lib_file_results).to have_successful_no_warnings_output
>       
>       expected: ["", "", 0]
>            got: ["", "", 5]
>       
>       (compared using ==)

Does anyone have any ideas for how to fix this?  I'll probably just tag these as `:failing_on_appveyor`.  I figure it isn't that important to run these specs on windows...I don't think Ruby's warnings are platform-specific, and allowing a whitelist of stdlibs isn't really a platform issue either.

lib/rspec/support/ruby_features.rb
```diff
@@ -1,3 +1,5 @@
+require 'rbconfig'
```
JonRowe: The irony of adding an additional piece of standard library while detecting usage of the standard library amuses me ;)

lib/rspec/support/spec/library_wide_checks.rb
```diff
@@ -0,0 +1,164 @@
...
+    @loaded_features_outfile = if ENV['CI']
+                                 # On AppVeyor we get exit status 5 ("Access is Denied",
+                                 # from what I've read) when trying to write to a tempfile.
+                                 #
+                                 # On Travis, we occasionally get Errno::ENOENT (No such file
+                                 # or directory) when reading from a tempfile.
+                                 #
+                                 # In both cases if we leave a file behind in the current dir,
+                                 # it's not a big deal so we put it in the current dir.
+                                 File.join(".", "loaded_features.txt")
```
JonRowe: I think this should just be `loaded_features.txt` on AppVeyor....

lib/rspec/support/ruby_features.rb
myronmarston: Well, I noticed that rspec-core loads rbconfig:

https://github.com/rspec/rspec-core/blob/v3.2.0/lib/rspec/core.rb#L5

...but it's not used in rspec-core anywhere.  Originally it was used for the OS detection that got moved here but the require never got moved.  It should be here since it is used here.  I will be removing it from rspec-core soon.

myronmarston:
> I think this should just be loaded_features.txt on AppVeyor....

Trying that now...

JonRowe:
Hmm, nope, maybe we should create the full path with `Dir.pwd`?

myronmarston:
> Hmm, nope, maybe we should create the full path with Dir.pwd?

I don't care enough to put effort into the slow trial-and-error feedback loop of AppVeyor.  I just marked them as `:failing_on_appveyor` and someone else who uses Windows can perhaps come along and fix them.  I'm not really concerned about the things these specs enforce differing on Windows, anyway.

JonRowe:
LGTM then...

PR #180, opened 2015-02-16
Fix warnings / stdlib spec on appveyor

JonRowe:
attempt to fix feature detection on appveyor

### DIFF
--- a/lib/rspec/support/spec/library_wide_checks.rb
+++ b/lib/rspec/support/spec/library_wide_checks.rb
@@ -46,12 +46,10 @@ RSpec.shared_examples_for "library wide checks" do |lib, options|
   define_method :load_all_lib_files do
     files = all_lib_files - lib_test_env_files
     preamble  = ['orig_loaded_features = $".dup', preamble_for_lib]
-    postamble = [
-      'loaded_features = ($" - orig_loaded_features).join("\n")',
-      "File.open('#{loaded_features_outfile}', 'w') { |f| f.write(loaded_features) }"
-    ]
+    postamble = ['puts(($" - orig_loaded_features).join("\n"))']
 
-    load_all_files(files, preamble, postamble)
+    @loaded_feature_lines, stderr, exitstatus = load_all_files(files, preamble, postamble)
+    ["", stderr, exitstatus]
   end
 
   define_method :load_all_spec_files do
@@ -60,11 +58,10 @@ RSpec.shared_examples_for "library wide checks" do |lib, options|
     load_all_files(files, preamble_for_spec)
   end
 
-  attr_reader :loaded_features_outfile, :all_lib_files, :lib_test_env_files,
+  attr_reader :all_lib_files, :lib_test_env_files,
               :lib_file_results, :spec_file_results
 
   before(:context) do
-    @loaded_features_outfile = File.join(".", "loaded_features.txt")
     @all_lib_files           = files_to_require_for("lib")
     @lib_test_env_files      = all_lib_files.grep(consider_a_test_env_file)
 
@@ -75,13 +72,11 @@ RSpec.shared_examples_for "library wide checks" do |lib, options|
     ].map(&:join).map(&:value)
   end
 
-  after(:context) { File.delete(loaded_features_outfile) }
-
   def have_successful_no_warnings_output
     eq ["", "", 0]
   end
 
-  it "issues no warnings when loaded", :slow, :failing_on_appveyor do
+  it "issues no warnings when loaded", :slow do
     expect(lib_file_results).to have_successful_no_warnings_output
   end
 
@@ -90,8 +85,8 @@ RSpec.shared_examples_for "library wide checks" do |lib, options|
   end
 
   it 'only loads a known set of stdlibs so gem authors are forced ' \
-     'to load libs they use to have passing specs', :slow, :failing_on_appveyor do
-    loaded_features = File.read(loaded_features_outfile).split("\n")
+     'to load libs they use to have passing specs', :slow do
+    loaded_features = @loaded_feature_lines.split("\n")
     if RUBY_VERSION == '1.8.7'
       # On 1.8.7, $" returns the relative require path if that was used
       # to require the file. LIB_REGEX will not match the relative version


JonRowe:
Ping @myronmarston this works on both Travis and Appveyor

lib/rspec/support/spec/library_wide_checks.rb
```diff
@@ -47,11 +47,11 @@ def load_all_files(files, preamble, postamble=nil)
     files = all_lib_files - lib_test_env_files
     preamble  = ['orig_loaded_features = $".dup', preamble_for_lib]
     postamble = [
-      'loaded_features = ($" - orig_loaded_features).join("\n")',
-      "File.open('#{loaded_features_outfile}', 'w') { |f| f.write(loaded_features) }"
+      'puts ($" - orig_loaded_features).join("\n")'
     ]
```
myronmarston: These 3 lines can be collapsed into one:

``` ruby
    postamble = ['puts ($" - orig_loaded_features).join("\n")']
```

lib/rspec/support/spec/library_wide_checks.rb
```diff
@@ -47,11 +47,11 @@ def load_all_files(files, preamble, postamble=nil)
...
 
-    load_all_files(files, preamble, postamble)
+    @loaded_features << load_all_files(files, preamble, postamble)[0]
+    load_all_files(files, [preamble_for_lib])
```
myronmarston: One of my goals in setting this up was to not load all lib files twice, which is what you are doing here. It's slow so it's preferrable to do it once and leverage it for both warnings and loaded features.  Here you're loading all lib files twice instead.  I think we can get by with doing it only once with:

``` ruby
    @loaded_features, *rest = load_all_files(files, preamble, postamble)
    return "", *rest
```

Basically, load it once and capture `@loaded_features` from stdout, and return `""` for stdout in its place to satisfy the expectation.  For the warnings detection, the warnings are on stderr anyway, and if there was a weird one that showed up on stdout our loaded feature check would catch it (since it requires a whitelisted regex for each line) so there's not really risk of making this a false positive by returning `""` for stdout.  Really, we probably don't even need to assert on stdout for the warnings specs....it just seemed simple enough to do it when I set it up.

lib/rspec/support/spec/library_wide_checks.rb
```diff
@@ -60,11 +60,11 @@ def load_all_files(files, preamble, postamble=nil)
     load_all_files(files, preamble_for_spec)
   end
 
-  attr_reader :loaded_features_outfile, :all_lib_files, :lib_test_env_files,
+  attr_reader :all_lib_files, :lib_test_env_files,
               :lib_file_results, :spec_file_results
 
   before(:context) do
-    @loaded_features_outfile = File.join(".", "loaded_features.txt")
+    @loaded_features         = ""
```
myronmarston: This line is unneeded.

lib/rspec/support/spec/library_wide_checks.rb
```diff
@@ -90,8 +88,8 @@ def have_successful_no_warnings_output
   end
 
   it 'only loads a known set of stdlibs so gem authors are forced ' \
-     'to load libs they use to have passing specs', :slow, :failing_on_appveyor do
-    loaded_features = File.read(loaded_features_outfile).split("\n")
+     'to load libs they use to have passing specs', :slow do
+    loaded_features = @loaded_features.split("\n")
```
myronmarston: It's a little weird to use "loaded_features" for the name of both variables when they are different (array vs string).  Maybe make the instance variable be `@loaded_features_lines` to make the difference more obvious?

myronmarston:
The approach seems sound enough.

JonRowe:
Hey @myronmarston I updated this.

JonRowe:
Any idea why the older versions of ruby are complaining?

lib/rspec/support/spec/library_wide_checks.rb
```diff
@@ -46,12 +46,10 @@ def load_all_files(files, preamble, postamble=nil)
...
     preamble  = ['orig_loaded_features = $".dup', preamble_for_lib]
-    postamble = [
-      'loaded_features = ($" - orig_loaded_features).join("\n")',
-      "File.open('#{loaded_features_outfile}', 'w') { |f| f.write(loaded_features) }"
-    ]
+    postamble = ['puts ($" - orig_loaded_features).join("\n")']
 
-    load_all_files(files, preamble, postamble)
+    @loaded_feature_lines, *rest = load_all_files(files, preamble, postamble)
+    [""] + rest
```
myronmarston: What's the point in returning `[""] + rest` rather than just `rest`?

lib/rspec/support/spec/library_wide_checks.rb
```diff
@@ -46,12 +46,10 @@ def load_all_files(files, preamble, postamble=nil)
   define_method :load_all_lib_files do
     files = all_lib_files - lib_test_env_files
     preamble  = ['orig_loaded_features = $".dup', preamble_for_lib]
-    postamble = [
-      'loaded_features = ($" - orig_loaded_features).join("\n")',
-      "File.open('#{loaded_features_outfile}', 'w') { |f| f.write(loaded_features) }"
-    ]
+    postamble = ['puts ($" - orig_loaded_features).join("\n")']
```
myronmarston: This is what's causing the travis failure on older rubies:

```
➜  ~  ruby -v
ruby 1.9.3p448 (2013-06-27 revision 41675) [x86_64-darwin12.4.0]
➜  ~  ruby -w -e "puts ('a' + 'b')"
-e:1: warning: (...) interpreted as grouped expression
ab
```

Ruby doesn't like you having a space between a method name and its arguments when you use parens.  I'd just switch back to the two-line form like we had before.

lib/rspec/support/spec/library_wide_checks.rb
JonRowe: This matches how it was before, and also your originally suggested replacement...

lib/rspec/support/spec/library_wide_checks.rb
```diff
@@ -46,12 +46,10 @@ def load_all_files(files, preamble, postamble=nil)
   define_method :load_all_lib_files do
     files = all_lib_files - lib_test_env_files
     preamble  = ['orig_loaded_features = $".dup', preamble_for_lib]
-    postamble = [
-      'loaded_features = ($" - orig_loaded_features).join("\n")',
-      "File.open('#{loaded_features_outfile}', 'w') { |f| f.write(loaded_features) }"
-    ]
+    postamble r ['puts(($" - orig_loaded_features).join("\n"))']
```
myronmarston: s/r/=/

myronmarston:
> > What's the point in returning [""] + rest rather than just rest?
> This matches how it was before, and also your originally suggested replacement...

Oh right, I forgot that it needs to return an array of 3 elements and we've pealed off the first.  What do you think about this instead?

``` ruby
@loaded_feature_lines, stderr, exitstatus = load_all_files(files, preamble, postamble)
return "", stderr, exitstatus
```

Or even this:

``` ruby
load_all_files(files, preamble, postamble).tap do |stdout, *|
  @loaded_feauture_lines = stdout.dup
  stdout.replace("")
end
```

IMO either one is much more clear what it's doing compared to the `*rest` and `[""] + rest` form.

JonRowe:
We use the implicit return format elsewhere here, so I'm just being consistent, but I don't mind returning specifically named variables

myronmarston:
> We use the implicit return format elsewhere here, so I'm just being consistent, but I don't mind returning specifically named variables

Implicit return is fine but I would like specifically named variables here.  I forgot that the return value is a tuple of exactly 3 values, not a general purpose array of a variable number of items, so the appending of `""` confused me.  Using explicit variables will make it more clear.

myronmarston:
Thanks, @JonRowe!

PR #183, opened 2015-02-18
Remove deprecated shared example group.

myronmarston:


JonRowe:
Appveyor has gone away for some reason but this LGTM so merging regardless.

PR #184, opened 2015-02-22
Move equality output formatting code out of rspec-expectations

gavingmiller:
Part of a larger effort originating from rspec/rspec-mocks#898 to pull Time, DateTime, and BigDecimal comparison code into a central location, and then apply across the entire rspec suite of projects.

### DIFF
--- a/spec/rspec/support/object_inspector_spec.rb
+++ b/spec/rspec/support/object_inspector_spec.rb
@@ -1,0 +1,97 @@
+require 'rspec/support/object_inspector'
+require 'rspec/matchers/fail_matchers'
+
+module RSpec
+  module Support
+    describe ObjectInspector, ".inspect" do
+      context 'with Time objects' do
+        let(:time) { Time.utc(1969, 12, 31, 19, 01, 40, 101) }
+        let(:formatted_time) { ObjectInspector.inspect(time) }
+
+        it 'produces an extended output' do
+          expected_output = "1969-12-31 19:01:40.000101"
+          expect(formatted_time).to include(expected_output)
+        end
+      end
+
+      context 'with DateTime objects' do
+        def with_date_loaded
+          in_sub_process_if_possible do
+            require 'date'
+            yield
+          end
+        end
+
+        let(:date_time) { DateTime.new(2000, 1, 1, 1, 1, Rational(1, 10)) }
+        let(:formatted_date_time) { ObjectInspector.inspect(date_time) }
+
+        it 'formats the DateTime using inspect' do
+          with_date_loaded do
+            expect(formatted_date_time).to eq(date_time.inspect)
+          end
+        end
+
+        it 'does not require DateTime to be defined since you need to require `date` to make it available' do
+          hide_const('DateTime')
+          expect(ObjectInspector.inspect('Test String')).to eq('"Test String"')
+        end
+
+        context 'when ActiveSupport is loaded' do
+          it "uses a custom format to ensure the output is different when DateTimes differ" do
+            stub_const("ActiveSupport", Module.new)
+
+            with_date_loaded do
+              expected_date_time = 'Sat, 01 Jan 2000 01:01:00.100000000 +0000'
+              expect(formatted_date_time).to eq(expected_date_time)
+            end
+          end
+        end
+      end
+
+      context 'with BigDecimal objects' do
+        let(:float)   { 3.3 }
+        let(:decimal) { BigDecimal("3.3") }
+
+        let(:formatted_decimal) { ObjectInspector.inspect(decimal) }
+
+        it 'fails with a conventional representation of the decimal' do
+          in_sub_process_if_possible do
+            require 'bigdecimal'
+            expect(formatted_decimal).to include('3.3 (#<BigDecimal')
+          end
+        end
+
+        it 'does not require BigDecimal to be defined since you need to require `bigdecimal` to make it available' do
+          hide_const('BigDecimal')
+          expect(ObjectInspector.inspect('Test String')).to eq('"Test String"')
+        end
+      end
+
+      context 'with objects that implement description' do
+        RSpec::Matchers.define :matcher_with_description do
+          match { true }
+          description { :description }
+        end
+
+        RSpec::Matchers.define :matcher_without_a_description do
+          match { true }
+          undef description
+        end
+
+        it "produces a description when a matcher object has a description" do
+          expect(ObjectInspector.inspect(matcher_with_description)).to eq(:description)
+        end
+
+        it "does not produce a description unless the object is a matcher" do
+          double = double('non-matcher double', :description => true)
+          expect(ObjectInspector.inspect(double)).to eq(double.inspect)
+        end
+
+        it "produces an inspected object when a matcher is missing a description" do
+          expect(ObjectInspector.inspect(matcher_without_a_description)).to eq(
+            matcher_without_a_description.inspect)
+        end
+      end
+    end
+  end
+end
--- a/lib/rspec/support/object_inspector.rb
+++ b/lib/rspec/support/object_inspector.rb
@@ -1,0 +1,48 @@
+module RSpec
+  module Support
+    # Provide additional output details beyond what `inspect` provides when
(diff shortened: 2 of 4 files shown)


spec/rspec/support/output_formatter_spec.rb
```diff
@@ -0,0 +1,107 @@
+require 'spec_helper'
+require 'rspec/support/output_formatter'
+require 'rspec/matchers/fail_matchers'
+
+module RSpec
+  module Support
+    describe OutputFormatter, ".format" do
```
myronmarston: None of the specs you have in this example group actually call `OutputFormatter.format`.  They are all just using the `eq` matcher logic that is still in rspec-expectations (but will later be refactored to use `OutputFormatter.format`).

The specs should be re-written to use `OutputFormatter.format`, particularly since that's the new code in this PR.

lib/rspec/support/output_formatter.rb
```diff
@@ -0,0 +1,46 @@
+module RSpec
+  module Support
+    # Provide additional output details beyond what `inspect` provides when
+    # printing Time, DateTime, or BigDecimal
+    module OutputFormatter
```
myronmarston: I think `OutputFormatter` is perhaps not the best name. Within the RSpec ecosystem, "formatter" has a pretty specific meaning of an object that implements rspec-core's formatter protocol and receives notification of events in order to format output.  This isn't like that at all.

Maybe `ObjectDescriber` or `ObjectInspector` is better?  The latter seems particularly appropriate since this is essentially a replacement for a less useful `inspect` call.

lib/rspec/support/output_formatter.rb
```diff
@@ -0,0 +1,46 @@
...
+        if Time === object
+          format_time(object)
+        elsif defined?(DateTime) && DateTime === object
+          format_date_time(object)
+        elsif defined?(BigDecimal) && BigDecimal === object
+          "#{object.to_s 'F'} (#{object.inspect})"
+        else
+          object.inspect
+        end
+      end
```
myronmarston: I think this method should also handle the case of calling `description` on an rspec matcher that implements `description`.  Currently we have this method for that:

https://github.com/rspec/rspec-support/blob/749e03b11ea78eaa643488328d1dc31f167af96f/lib/rspec/support/matcher_definition.rb#L34-L39

rspec-expectations has something similar:

https://github.com/rspec/rspec-expectations/blob/08c11431771b7755d2a402746b0fe4f8aec8d4ff/lib/rspec/matchers/composable.rb#L82-L85

Those both have a common purpose with this method in that they are trying to format an object for display in failure output.  Ideally, that would all be centralized in this one place so that all parts of RSpec benefit from common object formatting improvements.

spec/rspec/support/output_formatter_spec.rb
gavingmiller: You're totally right! :flushed:

myronmarston: > You're totally right!

It's always a good idea to see each spec go red at least once by writing it before the implementation or, if you write it afterwards, "breaking" the implementation temporarily.  It guards against this kind of issue by ensuring your tests are actually testing what you intend them to.

lib/rspec/support/output_formatter.rb
gavingmiller: Sounds good. Will work that into this PR.

spec/rspec/support/object_inspector_spec.rb
```diff
@@ -0,0 +1,74 @@
+require 'spec_helper'
+require 'rspec/support/object_inspector'
+require 'rspec/matchers/fail_matchers'
+
+module RSpec
+  module Support
+    describe ObjectInspector, ".inspect" do
+      include RSpec::Matchers::FailMatchers
```
myronmarston: I don't think you're using the fail matchers anymore, right?

spec/rspec/support/object_inspector_spec.rb
```diff
@@ -0,0 +1,74 @@
+require 'spec_helper'
```
myronmarston: Our `.rspec` file has a flag to require this automatically so this line isn't needed.

spec/rspec/support/object_inspector_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+          in_sub_process_if_possible do
+            require 'date'
+            yield
+          end
+        end
+
+        let(:date) { DateTime.new(2000, 1, 1, 1, 1, Rational(1, 10)) }
+        let(:formatted_date) { ObjectInspector.inspect(date) }
+
+        it 'produces a descript output using insepct' do
```
myronmarston: What does "a descript output" mean?  Also, s/insepct/inspect/.

spec/rspec/support/object_inspector_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+          expect(ObjectInspector.inspect('Test String')).to eq('"Test String"')
+        end
+
+        context 'when ActiveSupport is loaded' do
+          it "uses a custom format to ensure the output is different when DateTimes differ" do
+            stub_const("ActiveSupport", Module.new)
+
+            with_date_loaded do
+              expected_date = date.strftime("%a, %d %b %Y %H:%M:%S.%N %z")
+              expect(formatted_date).to eq(expected_date)
```
myronmarston: This expectation simply mirrors the implementation.  It'd be good to make `expected_date` be a known, static value.  Is that feasible?

spec/rspec/support/object_inspector_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+      context 'with DateTime objects' do
+        def with_date_loaded
+          in_sub_process_if_possible do
+            require 'date'
+            yield
+          end
+        end
+
+        let(:date) { DateTime.new(2000, 1, 1, 1, 1, Rational(1, 10)) }
+        let(:formatted_date) { ObjectInspector.inspect(date) }
```
myronmarston: I think it's a little confusing to call these `date` and `formatted_date` given you are dealing with a `DateTime` rather than a `Date` object.

myronmarston:
Much improved.  After you address my feedback above, please leave a comment letting us know. GitHub doesn't notify us when new commits are pushed to a PR so new commits could sit unreviewed for a long time if you don't make a point to let us know about them.

myronmarston:
As discussed in this commit:

https://github.com/rspec/rspec-support/pull/184#discussion_r25129580

...I think this still needs the handling for calling `description` on a matcher.

gavingmiller:
@myronmarston thanks for the feedback. I'm in the process of working out [#184 (comment)](https://github.com/rspec/rspec-support/pull/184#discussion_r25129580), I'll comment again once I've completed it.

gavingmiller:
@myronmarston moved the `description` code into the `ObjectInspector`. Locally tests pass, but looks like specs for the other libraries that depend on rspec-support are failing. What's the best way to go about taking care of those? I've not worked with a library with these types of dependencies before. Thanks!

myronmarston:
> looks like specs for the other libraries that depend on rspec-support are failing. What's the best way to go about taking care of those?

Ideally, you'd keep the old APIs (in this case, `rspec_description_for_object`) for now.  After this PR has been merged and PRs have been created and merged for the other repos moving them from the old API to the new one (`ObjectInspector`), you can open up a follow-up PR to this repo removing `rspec_description_for_object`.

gavingmiller:
I've incorporated the `description` on matcher into the `ObjectInspector`. This PR should be ready to go now. :smile:

myronmarston:
This looks great!  Can you squash it into one commit, @gavingmiller?  Or at least a small number of logical changesets?  I'll be happy to merge after that.

gavingmiller:
Done & Done. Thanks for all the help myron - it's a bit intimidating making a PR like this for a big project. :relaxed:

myronmarston:
> Done & Done. Thanks for all the help myron - it's a bit intimidating making a PR like this for a big project.

You're doing a great job...keep up the great work :).

PR #187, opened 2015-03-04
Remove `color` definition that is redefined later in the file.

myronmarston:
The `attr_reader :color` definition was only being used
so that we could define a `color?` alias, which seems kinda
silly.


JonRowe:
LGTM

PR #190, opened 2015-03-13
Object inspector now supports test double

sivagollapalli:


sivagollapalli:
@myronmarston It would great if you share your thoughts on this PR?

myronmarston:
> @myronmarston It would great if you share your thoughts on this PR?

Thanks for the reminder and sorry for the delay!  I plan to review this later today.  Feel free to ping me again if I don't get around to it by tomorrow.

lib/rspec/support/object_inspector.rb
```diff
@@ -13,6 +13,8 @@ def self.inspect(object)
           "#{object.to_s 'F'} (#{object.inspect})"
         elsif RSpec::Support.is_a_matcher?(object) && object.respond_to?(:description)
           object.description
+        elsif object.class == RSpec::Mocks::Double
```
JonRowe: Generally we would write this as `RSpec::Mocks::Double === object` which prevents the target tampering with the `==` operator on it's class (which is a possibility)

lib/rspec/support/object_inspector.rb
```diff
@@ -13,6 +13,8 @@ def self.inspect(object)
...
+          "Double (#{object.instance_variable_get(:@name)})"
```
JonRowe: Hmm I feel uncomfortable prying into the object like this, but exposing name directly would be worse...

lib/rspec/support/object_inspector.rb
myronmarston: We also can't count on `RSpec::Mocks::Double` being defined.  (For example, when rspec-expectations is used on a project but not rspec-mocks).  So this would have to be `defined?(RSpec::Mocks::Double) && RSpec::Mocks::Double === object`

lib/rspec/support/object_inspector.rb
myronmarston: Yeah, it's messy.  Here's an alternate idea, that addresses both this issue and the messiness of `defined?(RSpec::Mocks::Double)`: add an API like `ObjectInspector.register(RSpec::Mocks::Double) { ... }` so that rspec-mocks can register the inspect output for `RSpec::Mocks::Double`.  That way, rspec-support doesn't need to know anything about rspec-mocks -- rspec-mocks would simply register itself so that its objects are formatted well when failure messages use the `ObjectInspector`.

Thoughts?

sivagollapalli: @JonRowe I think we need to add `attr_reader :name` on `RSpec::Mocks::Double` to avoid prying into the object. Thoughts please?

myronmarston: > @JonRowe I think we need to add `attr_reader :name` on `RSpec::Mocks::Double` to avoid prying into the object. Thoughts please?

We intentionally have not exposed a `name` attribute off of `RSpec::Mocks::Double` because doubles are meant to be blank slate objects that have as few methods as possible initially defined so that the user is given free reign to define whatever interface they want w/o us encroaching on the interface w/ our own methods.  The user needs to be able to stub `name` for their own use.

So I think we need to pry into the double to get the instance variable, but doing so _from a completely separate library_ (rspec-support) feels wrong, and that's where my suggestion above helps: by providing an interface from rspec-support where rspec-mocks can register doubles with `ObjectInspector`, we can keep this logic in rspec-mocks where it's not nearly as big of a deal.

JonRowe: What @myronmarston said, we definitely do not want to use a plain `attr_reader` here.

sivagollapalli: @myronmarston As per your comment I understood following points:
 
1) We need to define a method called `ObjectInspector.register(double)` in `ObjectInspector` class.
2) The above method will be called from `rspec-mocks` to register doubles.
3)  Return value of above method would be `Double (Foo)`.

Please correct me if I understood anything wrong.

myronmarston: yep.

sivagollapalli:
@myronmarston PR updated as per discussion above.

lib/rspec/support/object_inspector.rb
```diff
@@ -13,11 +13,18 @@ def self.inspect(object)
...
         else
           object.inspect
         end
       end
 
+      def self.register(double)
+        raise "Please stub 'name' message on your double" unless double.respond_to?(:name)
+        "Double (#{double.name})"
+      end
```
myronmarston: This assumes that the argument will be an rspec-mocks test double.  The point of `register` is that it provides a way for other libraries like rspec-mocks to register how their objects will be presented by `ObjectInspector` w/o `ObjectInspector` having any knowledge of them.  So it should be something like:

``` ruby
def self.register(klass, &block)
  registered_klasses[klass] = block
end

def self.registered_klasses
  @registered_klasses ||= {}
end
```

Then in the `inspect` method above, in the `else` case (that currently has `object.inspect`), it should be something like:

``` ruby
registered_klasses.each do |klass, inspector|
  return inspector.call(object) if klass === object
end

object.inspect
```

In that way, `ObjectInspector` will have zero knowledge of rspec-mocks but exposes an interface that rspec-mocks can use to register how test doubles should be presented.  Does that make sense?

sivagollapalli:
@myronmarston Yes that makes sense. So I have updated PR as per your suggestion. But on travis one spec is failing. Do we the need that as `ObjectInspector` now supports doubles?

spec/rspec/support/object_inspector_spec.rb
```diff
@@ -92,6 +92,17 @@ def with_date_loaded
...
+      context "with double objects" do
+        it "displays double name" do
+          double = double('Foo')
+          allow(double).to receive(:name).and_return("Foo")
+          ObjectInspector.register double.class do |double|
+            "Double (#{double.name})"
+          end
+          expect(ObjectInspector.inspect(double)).to eq("Double (Foo)")
+        end
+      end
```
myronmarston: Rather than writing a spec for the exact case that's driving this (rspec-mocks test doubles) I think it's better to register a custom class and use that, for a few reasons:

* It proves that the register API works for _any_ arbitrary type, not just built-in rspec types.
* The registry is global state, which means that if you write a test here that registers inspect output for an `RSpec::Mocks::Double`, it will stomp the one that rspec-mocks will be registering once you work up that PR.
* What you've registered here works for this specific double (since you've stubbed `name`) but doesn't work for other doubles (which don't have `name` doubles).

So I'd recommend creating a custom type like `Currency` or `Person` and registering it, and using it in this spec.

We also want a spec showing that subclasses can use the registered inspect logic.

spec/rspec/support/object_inspector_spec.rb
```diff
@@ -92,6 +92,17 @@ def with_date_loaded
             matcher_without_a_description.inspect)
         end
       end
+
+      context "with double objects" do
+        it "displays double name" do
+          double = double('Foo')
+          allow(double).to receive(:name).and_return("Foo")
+          ObjectInspector.register double.class do |double|
```
JonRowe: FYI your specs are failing on Travis because the `double` here shadows / overwrites the existing local variable in this scope which is considered "less than ideal" practise and we have rubocop checking for it :)

JonRowe: FYI your specs are failing on Travis because the `double` here shadows / overwrites the existing local variable in this scope which is considered "less than ideal" practise and we have rubocop checking for it :)

sivagollapalli:
@myronmarston Updated the PR with specs.

spec/rspec/support/object_inspector_spec.rb
myronmarston: This spec should be removed.  As I tried to explain above, it's problematic for a few reasons and is really unnecessary since the implementation is blissfully unaware of rspec-mocks, test doubles, or any other such thing.

Instead, we'll wind up having some specs in rspec-mocks for this kind of thing once we do a PR there that builds on this one.

spec/rspec/support/object_inspector_spec.rb
```diff
@@ -92,6 +92,46 @@ def with_date_loaded
...
+        it "works with subclasses" do
+          class Admin < User
+          end
+
+          _admin = Admin.new("FooBar")
+          ObjectInspector.register Admin do |admin|
+            "Admin (#{admin.name})"
+          end
+          expect(ObjectInspector.inspect(_admin)).to eq("Admin (FooBar)")
+        end
```
myronmarston: This spec doesn't actually test what it needs to.  All classes are subclasses (of `Object` if not specified) so it's not actually interesting to register a subclass and test that the registered format logic is used.

What's of interest here is that `Admin`, as a subclass of `User`, should be formatted with the registered `User` logic, w/o the need to register `Admin`.  IOW, we want the format logic to apply to a class and any subclasses, w/o the need to register for each subclass.

This is important, because rspec-mocks has several different types of doubles and we don't want to have to register each subclass to have it formatted properly.

sivagollapalli: @myronmarston What I understood is, expected output of `ObjectInspector.inspect(_admin)` should be `User (FooBar)`. Am I correct?

myronmarston: Yes, after you remove `ObjectInspector.register Admin`.

sivagollapalli:
@myronmarston Removed `ObjectInspector.register Admin` from specs

spec/rspec/support/object_inspector_spec.rb
myronmarston: Any reason you've kept this spec around?

spec/rspec/support/object_inspector_spec.rb
```diff
@@ -92,6 +92,44 @@ def with_date_loaded
...
+          expect(ObjectInspector.inspect(user)).to eq("User (FooBar)")
+        end
+
+        it "works with subclasses" do
+          class Admin < User
+          end
+
+          admin = Admin.new("FooBar")
+          expect(ObjectInspector.inspect(admin)).to eq("User (FooBar)")
+        end
```
myronmarston: There's an ordering dependency in these specs -- this example will only pass if the prior `it "works with custom class"` one runs first (since it registers the superclass).

The `register` call should be moved into a `before(:context)` hook.

sivagollapalli: @myronmarston Removed dependency across specs.

myronmarston:
LGTM.  Do you mind squashing this into one commit?

myronmarston:
It looks like you reverted to the original implementation w/o the fixes we discussed....

sivagollapalli:
@myronmarston By mistake I have reset the changes. I have included all discussed changes in commit. Please check.

myronmarston:
Thanks, @sivagollapalli!

sivagollapalli:
@myronmarston Could we close this https://github.com/rspec/rspec-expectations/issues/688 ? If not what else needs to be done so that I would give a shot!.

myronmarston:
> @myronmarston Could we close this rspec/rspec-expectations#688 ? If not what else needs to be done so that I would give a shot!.

We can't close it yet because rspec-mocks test doubles don't display in failures how we would like.

The solution is to open an rspec-mocks PR that uses the new `ObjectInspector.register` API to register `RSpec::Mocks::TestDouble` so that they are formatted how we like.

sivagollapalli:
@myronmarston I just opened PR(https://github.com/rspec/rspec-mocks/pull/920) to uses the new `ObjectInspector.register`. Could you check the spec. If it's fine then I would proceed for implementation.

PR #191, opened 2015-03-16
Extend ObjectInspector and have Differ use it for Hashes.

hackling:
This PR updates how the ObjectInspector to handle the formatting inside of Arrays and Hashes.

It also updates the differ to use the ObjectInspector when it encounters a hash.
#### Example Test

``` ruby
      it 'x' do
        require 'bigdecimal'
        require 'active_support'
        expect(
          {
            BigDecimal('3') => nil,
            DateTime.new(2015,1,1) => nil,
            nil => {
              BigDecimal('4') => nil
            }
          }
        ).to eq ({
            nil => {
              BigDecimal('4') => nil
            }
        })
      end
```

#### Old Output
```
  1) RSpec::Support::Differ x
     Failure/Error: expect(

       expected: {nil=>{4.0 (#<BigDecimal:7fa2bc280918,'0.4E1',9(18)>)=>nil}}
            got: {3.0 (#<BigDecimal:7fa2bc280b70,'0.3E1',9(18)>)=>nil, Thu, 01 Jan 2015 00:00:00.000000000 +0000=>nil, ni
l=>{4.0 (#<BigDecimal:7fa2bc280aa8,'0.4E1',9(18)>)=>nil}}

       (compared using ==)

       Diff:
       @@ -1,2 +1,4 @@
       -nil => {#<BigDecimal:7fa2bc280918,'0.4E1',9(18)>=>nil},
       +nil => {#<BigDecimal:7fa2bc280aa8,'0.4E1',9(18)>=>nil},
       +#<BigDecimal:7fa2bc280b70,'0.3E1',9(18)> => nil,
       +#<DateTime: 2015-01-01T00:00:00+00:00 ((2457024j,0s,0n),+0s,2299161j)> => nil,
```

#### New
```
  1) RSpec::Support::Differ x
     Failure/Error: expect(

       expected: {nil=>{4.0 (#<BigDecimal:7f9ad4d6e9c8,'0.4E1',9(18)>)=>nil}}
            got: {3.0 (#<BigDecimal:7f9ad4d6eb58,'0.3E1',9(18)>)=>nil, Thu, 01 Jan 2015 00:00:00.000000000 +0000=>nil, ni
l=>{4.0 (#<BigDecimal:7f9ad4d6eab8,'0.4E1',9(18)>)=>nil}}

       (compared using ==)

       Diff:
       @@ -1,2 +1,4 @@
       -nil => {4.0 (#<BigDecimal:7f9ad4d6e9c8,'0.4E1',9(18)>) => nil},
       +nil => {4.0 (#<BigDecimal:7f9ad4d6eab8,'0.4E1',9(18)>) => nil},
       +3.0 (#<BigDecimal:7f9ad4d6eb58,'0.3E1',9(18)>) => nil,
       +Thu, 01 Jan 2015 00:00:00.000000000 +0000 => nil,
```

### DIFF
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -2,7 +2,7 @@ RSpec::Support.require_rspec_support 'encoded_string'
 RSpec::Support.require_rspec_support 'hunk_generator'
 
 require 'pp'
-require 'rspec/support/object_inspector'
+RSpec::Support.require_rspec_support "object_inspector"
 
 module RSpec
   module Support
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -175,18 +175,19 @@ module RSpec
         end.join
       end
 
-      OBJECT_INSPECTOR_OPTIONS = {
-        :nested => false,
-        :joiner => ",\n",
-        :braces => false,
-        :trailing_comma => true
-      }
-
       def object_to_string(object)
         object = @object_preparer.call(object)
         case object
         when Hash
-          ObjectInspector.inspect(object, OBJECT_INSPECTOR_OPTIONS)
+          formatted_hash = ObjectInspector.formatter(object)
+          formatted_hash.keys.sort_by { |k| k.to_s }.map do |key|
+            pp_key   = PP.singleline_pp(key, "")
+            pp_value = PP.singleline_pp(formatted_hash[key], "")
+
+            "#{pp_key} => #{pp_value},"
+          end.join("\n")
+        when Array
+          PP.pp(ObjectInspector.formatter(object), "")
         when String
           object =~ /\n/ ? object : object.inspect
         else
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -179,13 +179,7 @@ module RSpec
         object = @object_preparer.call(object)
         case object
         when Hash
-          formatted_hash = ObjectInspector.formatter(object)
-          formatted_hash.keys.sort_by { |k| k.to_s }.map do |key|
-            pp_key   = PP.singleline_pp(key, "")
-            pp_value = PP.singleline_pp(formatted_hash[key], "")
-
-            "#{pp_key} => #{pp_value},"
-          end.join("\n")
+          hash_to_string(object)
         when Array
           PP.pp(ObjectInspector.formatter(object), "")
         when String
@@ -193,6 +187,16 @@ module RSpec
         else
           PP.pp(object, "")
         end
+      end
+
+      def hash_to_string(hash)
+        formatted_hash = ObjectInspector.formatter(hash)
+        formatted_hash.keys.sort_by { |k| k.to_s }.map do |key|
+          pp_key   = PP.singleline_pp(key, "")
+          pp_value = PP.singleline_pp(formatted_hash[key], "")
+
+          "#{pp_key} => #{pp_value},"
+        end.join("\n")
       end
 
       def handle_encoding_errors(actual, expected)
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -2,6 +2,7 @@ RSpec::Support.require_rspec_support 'encoded_string'
 RSpec::Support.require_rspec_support 'hunk_generator'
 
 require 'pp'
+require 'rspec/support/object_inspector'
 
 module RSpec
   module Support
@@ -174,16 +175,18 @@ module RSpec
         end.join
       end
 
+      OBJECT_INSPECTOR_OPTIONS = {
+        :nested => false,
+        :joiner => ",\n",
+        :braces => false,
+        :trailing_comma => true
+      }
+
       def object_to_string(object)
         object = @object_preparer.call(object)
         case object
         when Hash
-          object.keys.sort_by { |k| k.to_s }.map do |key|
-            pp_key   = PP.singleline_pp(key, "")
-            pp_value = PP.singleline_pp(object[key], "")
-
-            "#{pp_key} => #{pp_value},"
-          end.join("\n")
+          ObjectInspector.inspect(object, OBJECT_INSPECTOR_OPTIONS)
         when String
           object =~ /\n/ ? object : object.inspect
         else
--- a/lib/rspec/support/object_inspector.rb
+++ b/lib/rspec/support/object_inspector.rb
@@ -3,14 +3,21 @@ module RSpec
     # Provide additional output details beyond what `inspect` provides when
     # printing Time, DateTime, or BigDecimal
     module ObjectInspector
+      DEFAULT_OPTIONS = {
+        :nested => false,
+        :joiner => ',',
+        :braces => true,
+        :trailing_comma => false
+      }
       # @api private
       # rubocop:disable CyclomaticComplexity
       # rubocop:disable MethodLength
-
       def self.inspect(object, options=DEFAULT_OPTIONS)
         case object
         when Time
           format_time(object)
+        when Hash
+          format_hash(object, options)
         else
           if defined?(DateTime) && DateTime === object
             format_date_time(object)
@@ -29,6 +36,19 @@ module RSpec
       # rubocop:enable CyclomaticComplexity
       # rubocop:enable MethodLength
 
+      # @private
+      def self.format_hash(object, options)
+        nested_options = { :braces => true, :joiner => ', ' }
+
+        pairs = object.sort_by { |k, _| k.to_s }.map do |key, value|
+          "#{inspect(key, nested_options)} => #{inspect(value, nested_options)}"
+        end
+
+        results = pairs.join(options[:joiner])
+        results = "{#{results}}" if options[:braces]
+        results += ',' if options[:trailing_comma]
+        results
+      end
 
       TIME_FORMAT = "%Y-%m-%d %H:%M:%S"
 
--- a/lib/rspec/support/object_inspector.rb
+++ b/lib/rspec/support/object_inspector.rb
@@ -46,7 +46,7 @@ module RSpec
 
         results = pairs.join(options[:joiner])
         results = "{#{results}}" if options[:braces]
-        results += ',' if options[:trailing_comma]
+        results << ',' if options[:trailing_comma]
         results
       end
 
--- a/lib/rspec/support/object_inspector.rb
+++ b/lib/rspec/support/object_inspector.rb
@@ -4,23 +4,31 @@ module RSpec
     # printing Time, DateTime, or BigDecimal
     module ObjectInspector
       # @api private
-      def self.inspect(object)
-        if Time === object
-          format_time(object)
-        elsif defined?(DateTime) && DateTime === object
-          format_date_time(object)
-        elsif defined?(BigDecimal) && BigDecimal === object
-          "#{object.to_s 'F'} (#{object.inspect})"
-        elsif RSpec::Support.is_a_matcher?(object) && object.respond_to?(:description)
-          object.description
-        else
-          registered_klasses.each do |klass, inspector|
-            return inspector.call(object) if klass === object
-          end
+      # rubocop:disable CyclomaticComplexity
+      # rubocop:disable MethodLength
 
-          object.inspect
+      def self.inspect(object, options=DEFAULT_OPTIONS)
+        case object
+        when Time
+          format_time(object)
+        else
+          if defined?(DateTime) && DateTime === object
+            format_date_time(object)
+          elsif defined?(BigDecimal) && BigDecimal === object
+            "#{object.to_s 'F'} (#{object.inspect})"
+          elsif RSpec::Support.is_a_matcher?(object) && object.respond_to?(:description)
+            object.description
+          else
+            registered_klasses.each do |klass, inspector|
+              return inspector.call(object) if klass === object
+            end
+            object.inspect
+          end
         end
       end
+      # rubocop:enable CyclomaticComplexity
+      # rubocop:enable MethodLength
+
 
       TIME_FORMAT = "%Y-%m-%d %H:%M:%S"
 
--- a/lib/rspec/support/object_inspector.rb
+++ b/lib/rspec/support/object_inspector.rb
@@ -3,21 +3,26 @@ module RSpec
     # Provide additional output details beyond what `inspect` provides when
     # printing Time, DateTime, or BigDecimal
     module ObjectInspector
-      DEFAULT_OPTIONS = {
-        :nested => false,
-        :joiner => ',',
-        :braces => true,
-        :trailing_comma => false
-      }
       # @api private
+      def self.inspect(initial)
+        formatted = formatter(initial)
+        if String === formatted || Symbol === formatted
+          formatted
+        else
+          formatted.inspect
+        end
+      end
+
       # rubocop:disable CyclomaticComplexity
       # rubocop:disable MethodLength
-      def self.inspect(object, options=DEFAULT_OPTIONS)
+      def self.formatter(object)
         case object
         when Time
           format_time(object)
+        when Array
+          ArrayInspector.inspect(object)
         when Hash
-          format_hash(object, options)
+          HashInspector.inspect(object)
         else
           if defined?(DateTime) && DateTime === object
             format_date_time(object)
@@ -29,26 +34,12 @@ module RSpec
             registered_klasses.each do |klass, inspector|
               return inspector.call(object) if klass === object
             end
-            object.inspect
+            object
           end
         end
       end
       # rubocop:enable CyclomaticComplexity
       # rubocop:enable MethodLength
-
-      # @private
-      def self.format_hash(object, options)
-        nested_options = { :braces => true, :joiner => ', ' }
-
-        pairs = object.sort_by { |k, _| k.to_s }.map do |key, value|
-          "#{inspect(key, nested_options)} => #{inspect(value, nested_options)}"
-        end
-
-        results = pairs.join(options[:joiner])
-        results = "{#{results}}" if options[:braces]
-        results << ',' if options[:trailing_comma]
-        results
-      end
 
       TIME_FORMAT = "%Y-%m-%d %H:%M:%S"
 
@@ -82,6 +73,25 @@ module RSpec
 
       def self.registered_klasses
         @registered_klasses ||= {}
+      end
+      class ArrayInspector
+        def self.inspect(array)
+          array.map do |element|
+            if Array === element
+              ArrayInspector.inspect(element)
+            elsif Hash === element
+              HashInspector.inspect(element)
+            else
+              ObjectInspector.formatter(element)
+            end
+          end
+        end
+      end
+
+      class HashInspector
+        def self.inspect(hash)
+          Hash[ArrayInspector.inspect(hash.to_a)]
+        end
       end
     end
   end
--- a/spec/rspec/support/differ_spec.rb
+++ b/spec/rspec/support/differ_spec.rb
@@ -279,12 +279,12 @@ EOD
         it "outputs unified diff message of two hashes with hashes inside them" do
           expected_diff = %Q{
 @@ -1,2 +1,2 @@
--"b" => {"key"=>"#{formatted_time}", "key1"=>"value"},
-+"c" => {"key"=>"#{formatted_time}", "key1"=>"value"},
+-"b" => {"key_1"=>"#{formatted_time}", "key_2"=>"value"},
++"c" => {"key_1"=>"#{formatted_time}", "key_2"=>"value"},
 }
 
-          left_side_hash = {'c' => {'key' => time, 'key1' => 'value'}}
-          right_side_hash = {'b' => {'key' => time, 'key1' => 'value'}}
+          left_side_hash = {'c' => {'key_1' => time, 'key_2' => 'value'}}
+          right_side_hash = {'b' => {'key_1' => time, 'key_2' => 'value'}}
           diff = differ.diff(left_side_hash, right_side_hash)
           expect(diff).to be_diffed_as(expected_diff)
         end
--- a/spec/rspec/support/differ_spec.rb
+++ b/spec/rspec/support/differ_spec.rb
@@ -263,13 +263,13 @@ EOD
 
       context 'when special-case objects are inside hashes' do
         let(:time) { Time.utc(1969, 12, 31, 19, 01, 40, 101) }
-        let(:formatted_time) { ObjectInspector.inspect(time) }
+        let(:formatted_time) { ObjectInspector.formatter(time) }
 
         it "outputs unified diff message of two hashes with Time object keys" do
           expected_diff = %Q{
 @@ -1,2 +1,2 @@
--#{formatted_time} => "b",
-+#{formatted_time} => "c",
+-"#{formatted_time}" => "b",
++"#{formatted_time}" => "c",
 }
 
           diff = differ.diff({ time => 'c'}, { time => 'b' })
@@ -279,13 +279,42 @@ EOD
         it "outputs unified diff message of two hashes with hashes inside them" do
           expected_diff = %Q{
 @@ -1,2 +1,2 @@
--"b" => {"key" => #{formatted_time}, "key1" => "value"},
-+"c" => {"key" => #{formatted_time}, "key1" => "value"},
+-"b" => {"key"=>"#{formatted_time}", "key1"=>"value"},
++"c" => {"key"=>"#{formatted_time}", "key1"=>"value"},
 }
 
           left_side_hash = {'c' => {'key' => time, 'key1' => 'value'}}
           right_side_hash = {'b' => {'key' => time, 'key1' => 'value'}}
           diff = differ.diff(left_side_hash, right_side_hash)
+          expect(diff).to be_diffed_as(expected_diff)
+        end
+      end
+
+      context 'when special-case objects are inside arrays' do
+        let(:time) { Time.utc(1969, 12, 31, 19, 01, 40, 101) }
+        let(:formatted_time) { ObjectInspector.formatter(time) }
+
+        it "outputs unified diff message of two arrays with Time object keys" do
+          expected_diff = %Q{
+@@ -1,2 +1,2 @@
+-["#{formatted_time}", "b"]
++["#{formatted_time}", "c"]
+}
+
+          diff = differ.diff([time, 'c'], [time, 'b'])
+          expect(diff).to be_diffed_as(expected_diff)
+        end
+
+        it "outputs unified diff message of two arrays with hashes inside them" do
+          expected_diff = %Q{
+@@ -1,2 +1,2 @@
+-[{"b"=>"#{formatted_time}"}, "c"]
++[{"a"=>"#{formatted_time}"}, "c"]
+}
+
+          left_side_array = [{'a' => time}, 'c']
+          right_side_array = [{'b' => time}, 'c']
+          diff = differ.diff(left_side_array, right_side_array)
           expect(diff).to be_diffed_as(expected_diff)
         end
       end
--- a/spec/rspec/support/differ_spec.rb
+++ b/spec/rspec/support/differ_spec.rb
@@ -261,6 +261,35 @@ EOD
           expect(diff).to be_diffed_as(expected_diff)
         end
 
+      context 'when special-case objects are inside hashes' do
+        let(:time) { Time.utc(1969, 12, 31, 19, 01, 40, 101) }
+        let(:formatted_time) { ObjectInspector.inspect(time) }
+
+        it "outputs unified diff message of two hashes with Time object keys" do
+          expected_diff = %Q{
+@@ -1,2 +1,2 @@
+-#{formatted_time} => "b",
++#{formatted_time} => "c",
+}
+
+          diff = differ.diff({ time => 'c'}, { time => 'b' })
+          expect(diff).to be_diffed_as(expected_diff)
+        end
+
+        it "outputs unified diff message of two hashes with hashes inside them" do
+          expected_diff = %Q{
+@@ -1,2 +1,2 @@
+-"b" => {"key" => #{formatted_time}, "key1" => "value"},
++"c" => {"key" => #{formatted_time}, "key1" => "value"},
+}
+
+          left_side_hash = {'c' => {'key' => time, 'key1' => 'value'}}
+          right_side_hash = {'b' => {'key' => time, 'key1' => 'value'}}
+          diff = differ.diff(left_side_hash, right_side_hash)
+          expect(diff).to be_diffed_as(expected_diff)
+        end
+      end
+
         it "outputs unified diff of multi line strings" do
           expected = "this is:\n  one string"
           actual   = "this is:\n  another string"
--- a/spec/rspec/support/object_inspector_spec.rb
+++ b/spec/rspec/support/object_inspector_spec.rb
@@ -33,7 +33,7 @@ module RSpec
 
         it 'does not require DateTime to be defined since you need to require `date` to make it available' do
           hide_const('DateTime')
-          expect(ObjectInspector.inspect('Test String')).to eq('"Test String"')
+          expect(ObjectInspector.inspect('Test String')).to eq('Test String')
         end
 
         context 'when ActiveSupport is loaded' do
@@ -63,7 +63,7 @@ module RSpec
 
         it 'does not require BigDecimal to be defined since you need to require `bigdecimal` to make it available' do
           hide_const('BigDecimal')
-          expect(ObjectInspector.inspect('Test String')).to eq('"Test String"')
+          expect(ObjectInspector.inspect('Test String')).to eq('Test String')
         end
       end
 
--- a/spec/rspec/support/object_inspector_spec.rb
+++ b/spec/rspec/support/object_inspector_spec.rb
@@ -80,6 +80,19 @@ module RSpec
         end
       end
 
+      context 'with Array Objects' do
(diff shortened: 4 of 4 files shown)


lib/rspec/support/differ.rb
```diff
@@ -2,11 +2,14 @@
 RSpec::Support.require_rspec_support 'hunk_generator'
 
 require 'pp'
+require 'rspec/support/object_inspector'
 
 module RSpec
   module Support
     # rubocop:disable ClassLength
     class Differ
+      include RSpec::Support::ObjectInspector
```
myronmarston: I don't think there's any reason to include this module.  It's not meant to be used as a mixin (and in fact defines no instance methods).

lib/rspec/support/object_inspector.rb
```diff
@@ -4,19 +4,39 @@ module Support
...
+
+      # @private
+      def self.format_hash(object, nested)
+        objects = object.keys.sort_by { |k| k.to_s }.map do |key|
+          formatted_key = ObjectInspector.inspect(key, true)
+          formatted_value = ObjectInspector.inspect(object[key], true)
+          "#{formatted_key} => #{formatted_value}#{',' unless nested}"
+        end
+
+        nested ? "{#{objects.join(', ')}}" : objects.join("\n")
```
myronmarston: It's not clear to me what the intention here is for the nesting vs not nesting case.  I looked for specs for these cases (hoping they would demonstrate the difference and the intended behavior) but I couldn't find any.  Can you add some specs that specifically demonstrate these cases and hopefully explain why this is needed?

lib/rspec/support/object_inspector.rb
```diff
@@ -4,19 +4,39 @@ module Support
...
-        elsif defined?(DateTime) && DateTime === object
+        when defined?(DateTime) && DateTime === object
           format_date_time(object)
-        elsif defined?(BigDecimal) && BigDecimal === object
+        when defined?(BigDecimal) && BigDecimal === object
           "#{object.to_s 'F'} (#{object.inspect})"
-        elsif RSpec::Support.is_a_matcher?(object) && object.respond_to?(:description)
+        when defined?(Hash) && Hash === object
+          format_hash(object, nested)
+        when defined?(Array) && Array === object
```
myronmarston: `Hash` and `Array` are always defined, so there's no need for the `defined?` checks on them. More generally, if we're going to switch to a `case` statement we can leverage that directly rather than using `===`:

``` ruby
case object
when Time
  format_time(object)
when Hash
  format_hash(object, nested)
when Array
  "[#{object.map { |element| ObjectInspector.inspect(element, true) }.join(', ')}]"
else
  if defined?(DateTime) && DateTime === object
    format_date_time(object)
  elsif defined?(BigDecimal) && BigDecimal === object
    "#{object.to_s 'F'} (#{object.inspect})"
  elsif RSpec::Support.is_a_matcher?(object) && object.respond_to?(:description)
    object.description
  else
    object.inspect
  end
end
```

lib/rspec/support/object_inspector.rb
```diff
@@ -4,19 +4,39 @@ module Support
...
+          "[#{object.map { |element| ObjectInspector.inspect(element, true) }.join(', ')}]"
```
myronmarston: This array formatting logic is complicated enough that I think I'd like to see it extracted into its own helper method.

spec/rspec/support/differ_spec.rb
```diff
@@ -261,6 +261,35 @@ def inspect; "<BrokenObject>"; end
...
 
+        it "outputs unified diff message of two hashes with BigDecimal keys" do
+          in_sub_process_if_possible do
+            require 'bigdecimal'
+            expect {
+              diff = differ.diff({ BigDecimal('3') => 'b'}, { 3 => 'b' })
+              if expect(diff).to match(/3.0 \(#<BigDecimal/)
+                raise StandardError, 'Passed Test'
+              end
+            }.to raise_error(StandardError, 'Passed Test')
```
myronmarston: Why are you using `expect` in a conditional, then raising an error, then expecting that error to get raised?  It seems quite convoluted compared to:

``` ruby
diff = differ.diff({ BigDecimal('3') => 'b'}, { 3 => 'b' })
expect(diff).to match(/3.0 \(#<BigDecimal/)
```

spec/rspec/support/differ_spec.rb
```diff
@@ -261,6 +261,35 @@ def inspect; "<BrokenObject>"; end
...
+          it "outputs unified diff message of two hashes with DateTime keys with the correct formatting" do
+            in_sub_process_if_possible do
+              require 'date'
+              stub_const("ActiveSupport", Module.new)
+              allow(date_time).to receive(:strftime).and_return("Timestamp")
+
+              expect {
+                diff = differ.diff({ date_time => 'b'}, { 3 => 'b' })
+                raise StandardError, 'Passed Test' if expect(diff).to match %r[Timestamp]
+              }.to raise_error(StandardError, 'Passed Test')
```
myronmarston: Same here -- the "raise to signal a passing test" pattern is confusing with no benefit that I can see.

spec/rspec/support/differ_spec.rb
```diff
@@ -261,6 +261,35 @@ def inspect; "<BrokenObject>"; end
...
+            expect {
+              diff = differ.diff({ BigDecimal('3') => 'b'}, { 3 => 'b' })
+              if expect(diff).to match(/3.0 \(#<BigDecimal/)
+                raise StandardError, 'Passed Test'
+              end
+            }.to raise_error(StandardError, 'Passed Test')
+          end
+        end
+
+        context 'with ActiveSupport' do
```
myronmarston: The differ knows nothing about activesupport and it doesn't seem appropriate to specifically call that out and test it here.  Isn't the behavioral change simply that it uses the `ObjectInspector` to format hashes?  Maybe we can do one example using `Time` objects in hashes (which were the original motivation for special formatting) rather than one BigDecimal and one AS example.

lib/rspec/support/differ.rb
```diff
@@ -178,12 +181,7 @@ def object_to_string(object)
         object = @object_preparer.call(object)
         case object
         when Hash
-          object.keys.sort_by { |k| k.to_s }.map do |key|
-            pp_key   = PP.singleline_pp(key, "")
-            pp_value = PP.singleline_pp(object[key], "")
-
-            "#{pp_key} => #{pp_value},"
-          end.join("\n")
+          ObjectInspector.inspect(object)
```
myronmarston: Are there other situations where we'd want the `ObjectInspector` to be used?  Maybe for arrays?  I wonder if we can do away with `pp` entirely?

spec/rspec/support/object_inspector_spec.rb
```diff
@@ -67,6 +67,30 @@ def with_date_loaded
...
+        let(:array) { [[BigDecimal('3.3')]] }
+        let(:formatted_array) { ObjectInspector.inspect(array) }
+
+        it 'recursively uses itself to inspect objects within it' do
+          in_sub_process_if_possible do
+            require 'bigdecimal'
+            expect(formatted_array).to include('[[3.3 (#<BigDecimal')
+          end
+        end
+      end
```
myronmarston: Both of these specs show only part of the output.  It's important that the full structure of the array/hash is formatted properly so I think I'd like to see `expect(formatted_hash|array).to eq(expected_output)` rather than just part of it.

Also, I think it would work better to use `Time` objects -- the output is more deterministic than with `bigdecimal` and doesn't require the special `in_sub_process_if_possible` handling.

spec/rspec/support/differ_spec.rb
hackling: I guess when I wrote this, I didn't quite fully understand how `in_sub_process_if_possible` worked, as was under the impression that it only applied to the contents of what was inside `expects` with a block.

I agree that switching these tests to use time objects might be a better option. I decided to contribute to this repo initially as I wanted to change how BigDecimals outputted, but because we are outputting the inspect block from them, it gets all finicky when trying to match on the objects id.

lib/rspec/support/differ.rb
hackling: I was going to cover arrays but they don't seem to be handled by the differ.

spec/rspec/support/object_inspector_spec.rb
hackling: Agree with this on all fronts.

hackling:
What is the preferred method of having me action all the comments raised above?
Are you fine with me leaving the commits above and adding additional ones to amend the issues raised? 
Or would you prefer fixup commits that you can review separately, and then rebase before merging?
Or simply just have me fixup, rebase and force push?

lib/rspec/support/differ.rb
myronmarston: The are just called out with specific code for arrays, but I do believe arrays are rendered by the differ at times -- presumably by `pp`.  I'd have to experiment to see in which situations they get rendered and diffed but it definitely happens.

myronmarston:
> Are you fine with me leaving the commits above and adding additional ones to amend the issues raised? 

You can leave whatever follow up comments you want :).

> Or would you prefer fixup commits that you can review separately, and then rebase before merging?

That would be great.  It's helpful to review just the diff of whatever fixes you push, and then yeah, we'll want it squashed down before merging.

hackling:
@myronmarston I've added a bunch of commits. I haven't applied them as fixups, but once you give me the all clear, I'll go back and restructure them all appropriately before we merge.

lib/rspec/support/differ.rb
hackling: I did a bit of an investigation and here's what I found:

The diff view is only provided on things containing newline characters in them.
Hashes are given newline characters as we map over the key/value pairs and join each of them by them.

`PP.pp` will also add newline characters of it's own accord, if the length exceeds 79 characters.

So an Array **will** appear in the diff view if it:
 - does not contain procs
 - does not contain all numbers
 - will have a total length of more than 79 characters when represented as a string
 - any of it's string elements contain a new line character.

Running this test:
``` ruby
    expected = ["poo", "thot", "that"]
    actual = ["poo\n", "thot", "that", "car"]
    expect(expected).to eq actual
```
Will provide this output in the differ:
``` patch
       @@ -1,5 +1,4 @@
       -poo\n
       +poo
        thot
        that
       -car
``` 

Moving forward, I'm not really sure what we want the behaviour to be.

I also played around with mapping over the elements in an Array and running them through the ObjectInspector, before passing the new array to PP.pp, but it gave me all sorts of issues surrounding double quoted strings and escaping slashes.

hackling:
The more I dig deeper into this, the more I think we should leave out expanding the differ and the ObjectInspector to handle arrays. 
It's going to require a bit of discussion on how we want to handle it, and then how we are going to go about reworking the Differ, ObjectInspector and HunkGenerator to accommodate.

spec/rspec/support/differ_spec.rb
```diff
@@ -261,6 +261,48 @@ def inspect; "<BrokenObject>"; end
...
+@@ -1,2 +1,2 @@
+-{"key" => #{formatted_time}, "key1" => "value"} => "b",
++{"key" => #{formatted_time}, "key1" => "value"} => "c",
+}
+
+          left_side_hash = {{'key' => time, 'key1' => 'value'} => 'c'}
+          right_side_hash = {{'key' => time, 'key1' => 'value'} => 'b'}
+          diff = differ.diff(left_side_hash, right_side_hash)
+          expect(diff).to be_diffed_as(expected_diff)
+        end
```
myronmarston: It seems odd/nonstandard to use a hash as a key inside a hash.  It's far more common to have nested hashes used as values inside a hash.  Does this spec cover any code paths that using the nested hash as a value instead wouldn't?  If not, I think it would be nice to change this to be closer to what typical user usage will be.

spec/rspec/support/differ_spec.rb
```diff
@@ -261,6 +261,48 @@ def inspect; "<BrokenObject>"; end
...
```
myronmarston: The description of this example says it is about arrays but I don't see any arrays.  Actually, this looks like a duplicate of the last spec.

spec/rspec/support/object_inspector_spec.rb
myronmarston: Are you still planning to address this?

lib/rspec/support/object_inspector.rb
myronmarston: I noticed that you added some tests of how the nesting interacts with the differ but it would also be nice to see how the nesting affects raw `ObjectInspector.inspect` output since that method is going to be called directly from rspec-expectations and rspec-mocks.

spec/rspec/support/differ_spec.rb
hackling: Yeah it is. Already made note of it above.

spec/rspec/support/object_inspector_spec.rb
hackling: I've fixed it for hashes, but as mentioned in my comments, I think the formatting of Arrays should be left for another ticket pending a discussion about the direction rspec-support wants to take in formatting them via the differ.

hackling:
@myronmarston Just wondering if there is something you'd like me to do before this gets another look?

The commits are over the shop, but I've left them that way so it's easier for you to review them, before I go back and reconstruct them properly, prior to merging.

myronmarston:
> @myronmarston Just wondering if there is something you'd like me to do before this gets another look?

You need to tell me it's ready for another look :).  GitHub doesn't notify when new commits are pushed, so I had no idea there were new commits here waiting for a look.  I probably won't have time to look tonight but will try to get to it in the next day or two.

spec/rspec/support/differ_spec.rb
```diff
@@ -261,6 +261,35 @@ def inspect; "<BrokenObject>"; end
           expect(diff).to be_diffed_as(expected_diff)
         end
 
+      context 'when special-case objects inside hashes' do
```
myronmarston: This doc string doesn't really make sense grammatically.  Should it be "with special-case objects inside hashes" or "when special-case objects are inside hashes"?

spec/rspec/support/object_inspector_spec.rb
```diff
@@ -67,6 +67,18 @@ def with_date_loaded
...
       end
 
+      context 'with Hash Objects' do
+        let(:time) { Time.utc(1969, 12, 31, 19, 01, 40, 101) }
+        let(:hash) { { 'key' => { 'inner_key' => time } } }
+        let(:formatted_hash) { ObjectInspector.inspect(hash) }
+        let(:expected_formatting) { /\"key\" => \{\"inner_key\" => 1969-12-31 19:01:40\.000101(000)? \+0000\},/ }
+
+        it 'recursively uses itself to inspect objects within it' do
+
```
myronmarston: This blank line seems unnecessary -- mind removing it?

lib/rspec/support/object_inspector.rb
```diff
@@ -4,19 +4,39 @@ module Support
...
+          else
+            object.inspect
+          end
         end
       end
+      # rubocop:enable CyclomaticComplexity
+      # rubocop:enable MethodLength
+
+      # @private
+      def self.format_hash(object, nested)
```
myronmarston: I think this method can be refactored to a simpler implementation:

``` ruby
def self.format_hash(object, nested)
  pairs = object.sort_by { |k, v| k.to_s }.map do |key, value|
    "#{inspect(key, true)} => #{inspect(value, true)}#{',' unless nested}"
  end

  nested ? "{#{pairs.join(', ')}}" : pairs.join("\n")
end
```

myronmarston: Actually, the logic doesn't seem quite right, either...consider the output I'm seeing:

```
irb(main):008:0> ObjectInspector.inspect({ :a => 5 })
=> ":a => 5,"
```

Notice the lack of curly braces and the extra trailing comma.

Actually, I'm realizing that we probably want it that way for the diff (so that first and last keys aren't diffed differently) but for normal inspection in a failure message outside a diff, I think we'd want the curly braces and no trailing comma.

myronmarston:
Sorry about the delay, @hackling.  This looking pretty good.  I think it's close to ready to merge, although there are a few comments above that I think should be addressed.  It also needs to be rebased against master (#190 was merged and changed the `ObjectInspector` which causes a merge conflict but it's pretty easy to resolve, I think).

Please ping me next time when you push new commits so I know to take a look.  Thanks!

lib/rspec/support/object_inspector.rb
hackling: @myronmarston I think I understand what your asking for, but if you could just confirm

```
irb(main):008:0> ObjectInspector.inspect({ :a => 5, :b => 6 })
=> "{:a => 5, :b => 6}"
```

When used inside a differ:
```
irb(main):008:0> ObjectInspector.inspect({ :a => 5, :b => 6 })
=> ":a => 5,\n:b => 6,"
```

myronmarston: Exactly.

hackling: I'm a bit unsure about adding another boolean here to accommodate so that the inspector has knowledge that it's being invoked inside a "differ".

myronmarston: How about this, then...create a `DifferObjectInspector` class that subclasses `ObjectInspector` and overrides the specific parts that are different for diffing.  You may also need to extract some `ObjectInspector` helper methods in order to provide extension points that the `DifferObjectInspector` can customize by overriding.

hackling:
@myronmarston The rebase got a bit gross, and as we were nearing the end, I opted to rebuild the commits and force push.

Ready for review.

lib/rspec/support/differ.rb
```diff
@@ -2,6 +2,7 @@
 RSpec::Support.require_rspec_support 'hunk_generator'
 
 require 'pp'
+require 'rspec/support/object_inspector'
```
myronmarston: This should use `RSpec::Support.require_rspec_support "object_inspector"`.  That uses `require_relative` on rubies that support that and normal `require` on rubies that don't.  We prefer `require_relative` when its available because it's always O(1) regardless of how large your load path where as `require` is O(N) based on the size of your load path.

lib/rspec/support/object_inspector.rb
```diff
@@ -3,23 +3,51 @@ module Support
...
+        when Time
           format_time(object)
-        elsif defined?(DateTime) && DateTime === object
-          format_date_time(object)
-        elsif defined?(BigDecimal) && BigDecimal === object
-          "#{object.to_s 'F'} (#{object.inspect})"
-        elsif RSpec::Support.is_a_matcher?(object) && object.respond_to?(:description)
-          object.description
+        when Hash
+          format_hash(object, options)
```
myronmarston: Didn't you have an earlier version of this that also formatted arrays?  I think we'll want the more detailed formatting for time/big decimal/etc when those are in arrays, too.

lib/rspec/support/object_inspector.rb
```diff
@@ -3,23 +3,51 @@ module Support
...
+      def self.format_hash(object, options)
+        nested_options = { :braces => true, :joiner => ', ' }
+
+        pairs = object.sort_by { |k, _| k.to_s }.map do |key, value|
+          "#{inspect(key, nested_options)} => #{inspect(value, nested_options)}"
         end
+
+        results = pairs.join(options[:joiner])
+        results = "{#{results}}" if options[:braces]
+        results += ',' if options[:trailing_comma]
```
myronmarston: I'd prefer you used `results << ',' if ...` here.  `+=` allocates a new string object where as `<<` doesn't, so the latter is going to perform better.

hackling: Wasn't aware you could even do this. I like it!

lib/rspec/support/object_inspector.rb
hackling: I initially had it in there, but I removed it out over the course of the pull request. 

I think the formatting inside the arrays needs to be tackled in a separate issue, because it's going to require some discussion prior.

See: https://github.com/rspec/rspec-support/pull/191#issuecomment-82731419

hackling: @myronmarston

The way arrays are handled is based on how `pp` works, and it gets quite finicky especially when the length of characters you want to display is over a certain number.

I believe that if we are going to move it into the `ObjectInspector`, we really shouldn't be trying to replicate the behaviour of `pp` inside it. However, doing that we will require some discussion about how we want to display the output.

I don't think that I should go down the path of writing a solution, without having a discussion about it first, as it will be based purely on my opinion of how the output should be handled.

myronmarston: I thought the uncertainty around arrays was how to diff them, not how to format them for non-diff cases.  What you had for non-diff cases seemed fine....it was just the `ObjectInspector` output of each item formatted like an array literal, right?  Did that cause problems for the differ?

hackling: @myronmarston This explanation might not be 100% correct;

`pp` does some stuff regarding formatting it's output, and the differ class seems to use some of the characters that it inserts in order to work out how to split up the array into "lines" before doing a diff on each line.

The output also goes a bit funny when total number of characters in an array that has been converted to a string, goes over a certain limit. 
If it's under a certain size, `pp` will output the whole array on one line. If it's over a certain size, it'll add `\n` to the end of each element and multiline it for you. I believe you can also control how many characters before it invokes this functionality with an environment variable (or something similar).

Anyway, in order to keep the differ working, either this functionality has to be recreated within the ObjectInspector, or the differ class gets a bit of a rework to work without expecting the string passed in to have already been formatted by `pp`.

I think its probably better for us to update the ObjectInspector AND the differ, to handle special objects inside arrays, in one pass, as I expect the differ class to get some changes, and the ObjectInspector to be extended with those changes in mind.

It would be much easier to implement if we just multilined every array when we output on the differ, but it makes sense for small arrays (that are easy to read), to be displayed on line.

I personally think it's much better for us leave the scope of this pull request to just deal with hashes, and create separate issue/pull-request that deals with arrays, in the near future.

myronmarston: > Anyway, in order to keep the differ working, either this functionality has to be recreated within the ObjectInspector, or the differ class gets a bit of a rework to work without expecting the string passed in to have already been formatted by pp.

I think there's a 3rd option you haven't mentioned: `ObjectInspector` can replace items with their inspected equivalents and then delegate to `pp` to format the entire array rather than re-implementing that.  Then we don't have to concern our selves with the pretty print array formatting (pp takes care of that for us!) and stick to replacing objects with their better-output equivalents.  That should be pretty doable, right?

Anyhow, I'm realizing that our needs for diffing are significantly different from our needs for general object inspecting.  I really think we should make separate objects for these two cases.  Then one can focus on formatting of primitives and data structures for use in error messages and one can focus on formatting for diffing (which could subclass or delegate to the other inspector).  Among other things, I think that would allow us to remove the plethora of formatting options the `ObjectInspector` now takes.  Those options worry me for a few reasons:

* The point of the `ObjectInspector` is to take care of the formatting details so the caller doesn't have to, but by making options be part of the API we're moving some of the responsibility back to the caller.
* The options seem to be focused on the hash case, but the caller could be passing some other kind of object, in which case the options seem extraneous and weird.
* As I understand it, the options are there to support the differ vs non-differ case...but the options support _many_ combinations of values.  And yet, I don't think we want to support many combinations of values -- there's just two valid combinations we want to support, right?  Reifying those two sets of options as their own classes that implement their own logic seems like a step in the right direction.

Anyhow, if we do split the differ inspector from the general inspector I think it would make even more sense to handle arrays in the general inspector as part of this PR.  (The differ inspector could perhaps be pushed off to another PR if that helps limit scope here but it need not be).

As far as array/hash formatting for the general/non-differ inspector goes, you may want to follow the pattern we used for `RSpec::Expectations::Composable#surface_descriptions_in`:

* That method simply walks the data structure and replaces the items in it as appropriate with `DescribableItem` objects. ([source](https://github.com/rspec/rspec-expectations/blob/v3.2.1/lib/rspec/matchers/composable.rb#L99-L115)).
* The describable items then implement `inspect` and `pretty_print` so that if the array or hash is inspected or pretty printed the items inside use the desired description strings. ([source](https://github.com/rspec/rspec-expectations/blob/v3.2.1/lib/rspec/matchers/composable.rb#L172-L180))

I see that as being analogous for what we want for the non-diff case: we want to replace how time objects are inspected when they are nested in arrays/hashes, but not worry about the formatting of the array or hash itself.  It's nice to let ruby take care of that simply by using `Array#inspect` or `Hash#inspect` with the items in the data structures suitably replaced.

Sorry if this takes the PR in a different direction then what you had in mind, but I do think it will simplify and improve things.

Thoughts?

JonRowe:
(this was updated @myronmarston)

lib/rspec/support/object_inspector.rb
hackling: @myronmarston I agree with the options being messy, I was just trying to adapt the class to handle the cases that you want.
I think having slightly different functionality for an inspector for the general output and the differ output is fine, so long as we can reuse similar patterns between the both, possibly share a class they both inherit from.

I'm a little confused here, as I don't know where the description method originates, so I don't know what it does in terms of output - https://github.com/rspec/rspec-expectations/blob/v3.2.1/lib/rspec/matchers/composable.rb#L174
I don't understand how it handles arrays in particular.

> we want to replace how time objects are inspected when they are nested in arrays/hashes, but not worry about the formatting of the array or hash itself. It's nice to let ruby take care of that simply by using Array#inspect or Hash#inspect with the items in the data structures suitably replaced.

I'm not sure this will work for use in the base of:
```
[
  { key => TimeObject }
]
```
It seems that using `surface_descriptions_in` will see it's an array object, and then do whatever the description method does to display it. Perhaps it'll be more clear to me once I know what `description` does.

To be perfectly honest, part of my reasoning for removing Arrays from this branch was de-scoping. I just wanted to have to stop monkey patching `BigDecimal` when working inside of some tests :P, and I didn't expect it this pull request to last as long as it has. I want to see this through to completion, I just find myself with less and less time to work on it at the moment (and my desire for the original feature also waning). 

---

The issue I'm having is I want a recursive solution so that it inspects the objects inside objects inside objects, however I think the issue I'm having is that I was trying to do it all inside one object, and therefore handle the formatting.

Having a look at the links you provided, I think this might actually solve the problem I'm having, as it ensures that `.inspect` is only called **once** per array or hash.

``` ruby
class ObjectInspector
  def self.inspect(item)
    case
    when Array === item
      ArrayInspector.inspect(item).inspect
    when Hash === item
      Hash[ArrayInspector.inspect(item.to_a)].inspect
    when #other options
    else
      #default case
    end
  end
end

class ArrayInspector
  def self.inspect(array)
    array.map do |element|
      if Array === element
        ArrayInspector.inspect(element)
      else
        ObjectInspector.inspect(element)
      end
    end
  end
end
```

The only issue I see here is that a Hash inside of an Array, will have both `.inspect` called on the hash, and then again once it's in the array, which I think will have weird formatting issues.

hackling: > The only issue I see here is that a Hash inside of an Array, will have both .inspect called on the hash, and then again once it's in the array, which I think will have weird formatting issues.

I'm open to suggestions on how to solve this without passing some information to one of the classes that lets it know that its being invoked inside of itself.

myronmarston: > To be perfectly honest, part of my reasoning for removing Arrays from this branch was de-scoping. I just wanted to have to stop monkey patching BigDecimal when working inside of some tests :P, and I didn't expect it this pull request to last as long as it has. I want to see this through to completion, I just find myself with less and less time to work on it at the moment (and my desire for the original feature also waning).

That's good feedback.  Sorry for turning this into more of a marathon than you intended :(.  As a contributor to open source I've always appreciated the chance maintainers provide me to address feedback but not everyone has the time or energy to do that.

Let's punt on arrays for now.  Before releasing this I'd at least like for us to move away from the options hash to having separate inspectors (one for differ, one for other callers).  Do you want to take that on before we merge?  Or should we merge this as-is and then I (or one of the other @rspec/rspec) can work on refactoring it?

hackling: Haven't got time to write a longer response at the moment, but I just wanted to say that I'm going have another crack at it today

myronmarston: Great.  Shout if you want one of us to take it across the finish line.

hackling:
@myronmarston 
Again, ignore the state of the commits, but I believe this solution gives us the best of both worlds.

We can still use the same `ObjectInspector` class, but just have slightly different usages of it that use the same code core. Custom formatting can be handled be the classes that need it, as demonstrated by the `Differ`.

I'm actually much happier with this solution than all of the previous ones proposed.

I'm not sure what the test failures are from on Ruby 1.8.7. Perhaps you might be able to shed some light?

lib/rspec/support/object_inspector.rb
```diff
@@ -4,23 +4,42 @@ module Support
     # printing Time, DateTime, or BigDecimal
     module ObjectInspector
       # @api private
-      def self.inspect(object)
-        if Time === object
+      def self.inspect(initial)
+        formatted = formatter(initial)
+        if String === formatted || Symbol === formatted
```
myronmarston: So I see that the `Symbol === formatted` check is needed to make [this test](https://github.com/rspec/rspec-support/blob/ad758245d098c24b9fd4bd0bb7a4054e66b574c2/spec/rspec/support/object_inspector_spec.rb#L81-L83) pass but I think we should enforce that only strings are treated this way here.  IMO, the test is buggy, as the matcher protocol states that `description` should return a string, not a symbol.  We should update the `description` method.

lib/rspec/support/object_inspector.rb
```diff
@@ -55,6 +74,25 @@ def self.register(klass, &block)
       def self.registered_klasses
         @registered_klasses ||= {}
       end
+      class ArrayInspector
```
myronmarston: Not a merge blocker, but a blank line before this class def would be nice.  Having it smashed together with the prior thing makes it harder to read.

lib/rspec/support/object_inspector.rb
```diff
@@ -55,6 +74,25 @@ def self.register(klass, &block)
...
+        def self.inspect(array)
+          array.map do |element|
+            if Array === element
+              ArrayInspector.inspect(element)
+            elsif Hash === element
+              HashInspector.inspect(element)
+            else
+              ObjectInspector.formatter(element)
+            end
+          end
```
myronmarston: Maybe I'm missing something, but can't this method def be simplified to:

``` ruby
def self.inspect(array)
  array.map do |element|
    ObjectInspector.formatter(element)
  end
end
```

After all, `ObjectInspector` takes care of the array and hash cases already.

myronmarston:
> I'm actually much happier with this solution than all of the previous ones proposed.

Me too.  This is great!

I left a few comments but don't consider them merge blockers -- I'm happy to address them myself post merge.  They're very minor.

> I'm not sure what the test failures are from on Ruby 1.8.7. Perhaps you might be able to shed some light?

Oh the joys of supporting so many different ruby interpreters and versions....

Looking into that now.

myronmarston:
The 1.8.7 failure is happening in rspec-mocks (we have the travis build configured to run the spec suites for the other rspec projects to ensure we haven't broken them by changes here).  I can't repro it locally, though.  I suspect it's an unrelated flapping test due to 1.8.7 weirdness that we hadn't noticed before.  I kicked the 1.8.7 build in the hope it'll pass now.

myronmarston:
...and it failed again :(.  Can you repro when you run the rspec-mocks spec suite against your branch of rspec-support?

lib/rspec/support/object_inspector.rb
myronmarston: I just discovered another problem with this (while trying this branch against rspec/rspec-expectations#740, which updates that gem to use `RSpec::Support::ObjectInspector.inspect`).  `RSpec::Support::ObjectInspector.inspect(obj)` is intended to be usable in place of `obj.inspect`, but due to this conditional, it behaves differently for strings compared to `String#inspect`.  `String#inspect` prints double-quote marks around the string, where as your changes here cause the string to be printed without the double-quote marks since it skips the `inspect` call.  As a result, it causes many failures in rspec-expectations.

myronmarston:
One other issue I'm seeing when running this against rspec/rspec-expectations#740.  I'm getting failures like this:

```
  1) yield_with_args matcher expect {...}.to yield_with_args(matcher, matcher) provides a description
     Failure/Error: expect(description).to eq("yield with args(a value within 0.2 of 1, a string matching /foo/)")

       expected: "yield with args(a value within 0.2 of 1, a string matching /foo/)"
            got: "yield with args(\"a value within 0.2 of 1\", \"a string matching /foo/\")"

       (compared using ==)
```
Notice that there are extra double quotes marks (essentially the opposite of the problem with strings I mentioned above).  The "a value within 0.2 of 1" bit is the description of a matcher, and as such isn't intended to look like a string literal, so the quote marks should not be included, but your changes here cause that.

I'm realizing that the changes you're attempting to make here, while they initially seem pretty straightforward, are actually pretty complicated to get right, once we start using the inspector defined here everywhere in rspec-expectations.  It's probably going to be pretty challenging to get it to work completely right.  You mentioned above that your time is limited, so let me know if you reach a point where you'd like one of us to take this PR across the finish line.  Your contribution is valued and your work here definitely makes RSpec better, but we do need to make sure it integrates properly and doesn't cause any regressions so we want to make sure we've got this right before we merge it.

Thanks!

myronmarston:
I'm taking this on now to get it across the finish line so we can include it in 3.3. Thanks for taking it this far, @hackling!

myronmarston:
Superseded by #191.

PR #193, opened 2015-04-10
Prefer calling `expected == actual` over `actual == expected`.

myronmarston:
This ensures that rspec-mocks' `anything` will always match,
even against objects that implement `==` wrongly.

### DIFF
--- a/spec/rspec/support/fuzzy_matcher_spec.rb
+++ b/spec/rspec/support/fuzzy_matcher_spec.rb
@@ -56,6 +56,37 @@ module RSpec
         end
       end
 
+      context "when given an object whose implementation of `==` wrongly assumes it will only be called with objects of the same type" do
+        Color = Struct.new(:r, :g, :b) do
+          def ==(other)
+            other.r == r && other.g == g && other.b == b
+          end
+        end
+
+        before(:context) do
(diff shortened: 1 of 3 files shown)


spec/rspec/support/fuzzy_matcher_spec.rb
```diff
@@ -56,6 +56,37 @@ module Support
...
+            other.r == r && other.g == g && other.b == b
+          end
+        end
+
+        before(:context) do
+          expect { Color.new(0, 0, 0) == Object.new }.to raise_error(NoMethodError, /undefined method `r'/)
+        end
+
+        it 'can match against an expected value that matches anything' do
+          anything = Object.new.tap do |o|
```
samphippen: This could just be `lambda { |*_| true }` right? Do you think that's better?

samphippen:
One comment, otherwise good to merge.

spec/rspec/support/fuzzy_matcher_spec.rb
myronmarston: I thought about using that, but `Lambda#===` was new in 1.9 so this doesn't work on 1.8.7.

PR #194, opened 2015-04-12
Add `RSpec::Support.failure_notifier`.

myronmarston:
This is intended to support the new `aggregate_failures`
feature discussed in rspec/rspec-expectations#733. I
put the `failure_notifier` here so that rspec-mocks can
use it as well -- that way, when we are aggregating failures,
we can aggregate rspec-mocks failures as well.

### DIFF
--- a/lib/rspec/support.rb
+++ b/lib/rspec/support.rb
@@ -73,6 +73,31 @@ module RSpec
       end
     end
 
+    # A single thread local variable so we don't excessively pollute that namespace.
+    def self.thread_local_data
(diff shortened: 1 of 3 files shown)


JonRowe:
Nice so we'll switch to using this instead of raising errors? Looks good to me, but needs a change log?

myronmarston:
This isn't a public API so it doesn't need a changelog.  The changelog will go in rspec-expectations and rspec-core for the new features this enables there.

PR #196, opened 2015-04-16
Demonstrate issue with large nokogiri documents

JonRowe:
Large nokogiri documents expand to even larger when pp'd, and this produces very complex diffs that take a long time to produce, I'm not sure if we have a solution for this but it's the source of rspec/rspec-expectations#773


benoittgt:
@JonRowe I don't see any activities on this issue since 4 years. I am in favor of closing it. What do you think?

JonRowe:
@benoittgt this is a bit confusing because of the rebase, but its worth leaving open as its an issue rather than a stale PR

PR #197, opened 2015-04-17
Don’t bother with coverage on windows.

myronmarston:
My rspec-mocks build (for rspec/rspec-mocks#932) is
failing on AppVeyor because the “issues no warnings
when spec files are loaded” check is getting stdout/stderr
output from simplecov, in spite of the fact that we set
the NO_COVERAGE ENV var when we shell out for that.
I’m not sure why it doesn’t work on Windows but I don’t
care to spend a bunch of time to figure it out, given
I don’t have access to a windows box.


myronmarston:
Merged so that hopefully rspec/rspec-mocks#932 can go green.

PR #198, opened 2015-04-20
Stop using `example.fail`.

myronmarston:
- Originally, this was equivalent to `raise`.
- Since we have added `RSpec::Matchers::FailMatchers`,
  in spec suites that include that module, `fail` now
  returns an rspec matcher rather than raising.
- It’s simpler just to raise.


JonRowe:
So this LGTM and the build failure is actually nokogiri failing to install on 1.9.3 so I feel it's safe to ignore, merging.

PR #199, opened 2015-04-27
Lazily produce the first non rspec backtrace line 

JonRowe:
Use exceptions to lazily find the backtrace line when `caller_locations` isn't available. A WIP for #937

### DIFF
--- a/lib/rspec/support/caller_filter.rb
+++ b/lib/rspec/support/caller_filter.rb
@@ -73,10 +73,24 @@ module RSpec
         end
       end
     else
+      CallerFinder = Class.new(StandardError)
+      class LazyBacktrace
+        def initialize(exception)
+          @exception = exception
+        end
+
+        def to_s
+          @exception.backtrace.find { |line| line !~ IGNORE_REGEX }
+        end
+      end
       # Earlier rubies do not support the two argument form of `caller`. This
       # fallback is logically the same, but slower.
       def self.first_non_rspec_line(*)
-        caller.find { |line| line !~ IGNORE_REGEX }
+        begin
+          raise CallerFinder
+        rescue CallerFinder => error
+          return LazyBacktrace.new(error)
+        end
       end
     end
   end
--- a/lib/rspec/support/caller_filter.rb
+++ b/lib/rspec/support/caller_filter.rb
@@ -27,7 +27,10 @@ module RSpec
     # with this complexity in our `RSpec.deprecate` calls, so we ignore it here.
     IGNORE_REGEX = Regexp.union(LIB_REGEX, "rubygems/core_ext/kernel_require.rb")
 
-    if RSpec::Support::RubyFeatures.caller_locations_supported?
+    # @private
+    CallerFinder = Class.new(StandardError)
+
+    if false #RSpec::Support::RubyFeatures.caller_locations_supported?
       # This supports args because it's more efficient when the caller specifies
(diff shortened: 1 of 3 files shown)


lib/rspec/support/caller_filter.rb
```diff
@@ -73,10 +73,24 @@ def self.first_non_rspec_line(skip_frames=3, increment=5)
...
+      end
       # Earlier rubies do not support the two argument form of `caller`. This
       # fallback is logically the same, but slower.
       def self.first_non_rspec_line(*)
-        caller.find { |line| line !~ IGNORE_REGEX }
+        begin
+          raise CallerFinder
+        rescue CallerFinder => error
+          return LazyBacktrace.new(error)
+        end
```
myronmarston: As discussed in rspec/rspec-mocks#937, `raise` is modestly _slower_ than `caller` when the backtrace is actually used.  So I don't think we want to always use `raise`...instead, it should be something the caller of `CallerFilter` opts in to due to knowing that the backtrace may not be used.  For situations where we know the backtrace may be used it should use the mechanism it's using now.

WDYT of `RSpec::CallerFilter.first_non_rspec_line` vs `RSpec::LazyCallerFilter.first_non_rspec_line`?

lib/rspec/support/caller_filter.rb
```diff
@@ -73,10 +73,24 @@ def self.first_non_rspec_line(skip_frames=3, increment=5)
...
       end
     else
+      CallerFinder = Class.new(StandardError)
+      class LazyBacktrace
+        def initialize(exception)
+          @exception = exception
+        end
+
+        def to_s
+          @exception.backtrace.find { |line| line !~ IGNORE_REGEX }
```
myronmarston: This could benefit from being memoized.

myronmarston:
The `to_s` thing is clever here; nice job.  Is it worth defining `to_str` as well?

Also, as discussed in rspec/rspec-mocks#937, I think we should also support this for rubies that provide `caller_filter` because I believe it'll still be significantly faster for cases where the backtrace may not be used.

And do you mind adding a benchmark to this repo as part of this PR?  (Feel free to steel the one I posted in that issue).

lib/rspec/support/caller_filter.rb
JonRowe: I've just benchmarked `raise` against `caller` and `caller_location` and added that to this PR and it seems that although a bare `raise` and capture is faster than both a bare `caller` or a bare `caller_location`, it isn't faster than the skip frames example for `caller_locations`?

myronmarston: > I've just benchmarked raise against caller and caller_location and added that to this PR and it seems that although a bare raise and capture is faster than both a bare caller or a bare caller_location, it isn't faster than the skip frames example for caller_locations?

Looks like you're right.  This surprises me!  At first I thought your benchmark was flawed in that the backtrace sizes were so small and with a larger, more realistic trace (e.g. 50 or 100 frames) it would be faster to use `raise`.  Turns out I was wrong -- so the commit I just pushed.

Given that all non-EOL'd versions of ruby have `caller_locations`, I think we shouldn't make a change here.  It's not worth it to me to add complexity to optimize EOL'd versions of ruby, which is all we can do here.  I do think it's worth pushing the benchmark to `master` so we have the record of it in the future.

JonRowe: Copy that, I'll extract the benchmark and push it to master

PR #202, opened 2015-05-04
Fix race condition in mkdir_p

fwiesel:
This occurs quite often with parallel_rspec. Consider following sequence of the two processes P1, P2:

P1: directory_exists?('reports') -> false
P2: directory_exists?('reports') -> false
P2: Dir.mkdir('reports') -> true
P1: Dir.mkdir('reports') -> Errno:EEXIST: `mkdir': File exists - ./spec/reports


JonRowe:
Seems a reasonable check, thanks. Merged despite build failure because it's down to "that gem" on 1.9.3. and bundler on appveyor.

PR #203, opened 2015-05-05
Attempt to fix appveyor

JonRowe:
Remove previous fixation of bundler on appveyor


myronmarston:
The failure we're seeing is for bundler 1.9.5, which makes me think this wasn't working anyway, perhaps?

JonRowe:
Well it passed... so.... *shrug*

PR #205, opened 2015-05-23
Improve inspector

myronmarston:
This builds off of #191 to address rspec/rspec-expectations#611.

@hackling--I used a slightly different solution for the `ObjectFormatter` but the differ changes are all yours.  Thanks again for contributing and for your patience as we get it integrated.


JonRowe:
LGTM :)

PR #206, opened 2015-05-24
Followup performance benchmarking for #205

JonRowe:
Given the surprise performance increase of using inject over map for
hash, I wanted to see if we could get the same performance boost with
the array part of the differ, but benchmarks show this is actually
slower.

More of a FYI to future development than anything else.

### DIFF
--- a/benchmarks/map_array.rb
+++ b/benchmarks/map_array.rb
@@ -1,0 +1,70 @@
+require 'benchmark/ips'
+
+def use_map_and_array_bracket(input)
+  input.map { |v| v.to_s }
+end
+
+def use_inject(input)
+  input.inject([]) do |array, v|
+    array << v.to_s
+    array
+  end
+end
+
+[10, 100, 1000].each do |size|
+  array = 1.upto(size)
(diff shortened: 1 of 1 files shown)


myronmarston:
> Given the surprise performance increase of using inject over map for
> hash, I wanted to see if we could get the same performance boost with
> the array part of the differ, but benchmarks show this is actually
> slower.

FWIW, I wasn't surprised that `inject` was faster for #205, and I'm not surprised that `map` is faster than `inject` here.

In #205, the `map` solution created an extra array allocation for every single pair in the input hash, which creates lots of extra garbage that needs to be GC'd, plus it had an extra `Hash.[]` call.  Mutating a hash in an `inject` block is kinda more natural solution there, and avoids the extra array allocations.

Here, you're benching `inject` against the kind of thing `map` was built for: applying a function (`to_s`) to every element in an array and returning a new array containing the results.  `inject` isn't going to be able to beat `map` for the kind of use case `map` is specifically meant for.

Given that this fits with my intuition, I personally find it to be "noise" in our benchmarks directory but if you still think it's valuable and want to keep it, merge away.

PR #208, opened 2015-05-27
[WIP] spike showing failure notifier with options

JonRowe:
The idea would be that normally you'd just set the failure notifier to something like:

```Ruby
RSpec::Support.failure_notifier = proc { |failure| method(:raise)(failure) }
```

but that `aggregate_failures` would change to something like:

```Ruby
def call(failure, options = {})
  failure.set_backtrace(caller) unless failure.backtrace
  failures << failure unless options[:eager]
end
```

### DIFF
--- a/lib/rspec/support.rb
+++ b/lib/rspec/support.rb
@@ -83,11 +83,11 @@ module RSpec
     end
 
     def self.failure_notifier
-      thread_local_data[:failure_notifier] ||= method(:raise)
+      thread_local_data[:failure_notifier] ||= lambda { |failure, _opts| method(:raise)(failure) }
     end
 
-    def self.notify_failure(failure)
-      failure_notifier.call(failure)
+    def self.notify_failure(failure, opts = {})
+      failure_notifier.call(failure, opts)
     end
 
     def self.with_failure_notifier(callable)


JonRowe:
^^ @myronmarston

lib/rspec/support.rb
```diff
@@ -83,11 +83,11 @@ def self.failure_notifier=(callable)
     end
 
     def self.failure_notifier
-      thread_local_data[:failure_notifier] ||= method(:raise)
+      thread_local_data[:failure_notifier] ||= lambda { |failure, _opts| method(:raise)(failure) }
```
myronmarston: This isn't valid syntax...

And actually, there's no reason to capture the `raise` method only to call it right away -- we can just send the `raise` method directly.  This could be:

``` ruby

    DEFAULT_FAILURE_NOTIFIER = lambda { |failure, _opts| raise failure }

    def self.failure_notifier
      thread_local_data[:failure_notifier] ||= DEFAULT_FAILURE_NOTIFIER
    end
```

lib/rspec/support.rb
```diff
@@ -83,11 +83,11 @@ def self.failure_notifier=(callable)
...
     end
 
-    def self.notify_failure(failure)
-      failure_notifier.call(failure)
+    def self.notify_failure(failure, opts = {})
+      failure_notifier.call(failure, opts)
```
myronmarston: Switching to the 2-argument form here breaks the other repos.  In the interest in having some green PRs we can merge individually rather than a set of PRs that are all independent, I think this should be something like:

``` ruby
def self.notify_failure(failure, opts = {})
  if failure_notifier.method(:call).arity == 1
    failure_notifier.call(failure)
  elsif Method === failure_notifier && failure_notifier.name == :raise
    # `raise` accepts 2 arguments (exception class and message) but we
    # don't want it to treat the opts hash as the message.
    failure_notifier.call(failure)
  else
    failure_notifier.call(failure, opts)
  end
end
```

...where the first two branches are temporary just to keep things passing, with the plan to remove them in a followup PR later once the other repos have been updated.

Note, however, that this breaks [this cuke](https://github.com/rspec/rspec-core/blob/master/features/expectation_framework_integration/aggregating_failures.feature) and I'm not yet sure why, so we'd want to look into that before merging (even though it won't be run as part of the build here and the build will likely to be green without us fixing that first).

myronmarston:
Closing in favor of #210.  Figured you're focused on the rspec-core profiling issue so I'd move this forward.

PR #212, opened 2015-06-03
Remove unneeded fallback logic.

myronmarston:
It was only temporarily needed to be backwards
compatible with the old one-arg notifier interface
that the other repos were using.


JonRowe:
Seems the specs where using the fallback logic :)

myronmarston:
Kicked the travis build now that rspec/rspec-mocks#962 is merged.

JonRowe:
LGTM

PR #215, opened 2015-06-16
Include delegators in inspection output

ahorner:
In current spec output, it’s not immediately obvious when an object
under test is actually wrapped in a delegator of some sort, which can 
lead to some tricky-to-track-down failures on equality checks for two 
seemingly-identical objects.

The fix proposed here is to toss objects into a special wrapper when we 
prepare them for inspection. This wrapper recursively modifies the 
object’s inspection string with delegator class information, which
should make it much more apparent when the object under test is wrapped
by a delegator of some sort.

Related to https://github.com/rspec/rspec-core/issues/1995.

### DIFF
--- a/lib/rspec/support/object_formatter.rb
+++ b/lib/rspec/support/object_formatter.rb
@@ -36,7 +36,7 @@ module RSpec
           elsif RSpec::Support.is_a_matcher?(object) && object.respond_to?(:description)
             inspection = object.description
           else
-            return object
+            return DelegatingInspector.new(object)
           end
         end
 
@@ -86,6 +86,21 @@ module RSpec
 
         def pretty_print(pp)
           pp.text inspection
+        end
+      end
+
+      # @private
+      DelegatingInspector = Struct.new(:object) do
+        def inspect
+          if defined?(::Delegator) && ::Delegator === object
+            "#<#{object.class}(#{ObjectFormatter.format(object.__getobj__)})>"
+          else
+            object.inspect
+          end
+        end
+
+        def pretty_print(pp)
+          pp.text inspect
         end
       end
     end
--- a/spec/rspec/support/object_formatter_spec.rb
+++ b/spec/rspec/support/object_formatter_spec.rb
@@ -95,6 +95,44 @@ module RSpec
         end
       end
 
+      context 'given a delegator' do
+        def with_delegate_loaded
+          in_sub_process_if_possible do
+            require 'delegate'
+            yield
+          end
+        end
+
+        let(:object) { Object.new }
+        let(:delegator) do
+          SimpleDelegator.new(object)
+        end
+
+        it 'includes the delegator class in the description' do
+          with_delegate_loaded do
+            expect(ObjectFormatter.format(delegator)).to eq "#<SimpleDelegator(#{object.inspect})>"
+          end
+        end
+
+        it 'does not require Delegator to be defined' do
+          hide_const("Delegator")
+          expect(ObjectFormatter.format(object)).to eq object.inspect
+        end
+
+        context 'for a specially-formatted object' do
+          let(:decimal) { BigDecimal("3.3") }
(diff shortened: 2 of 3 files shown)


ahorner:
Ah, this runs into an infinite recursion issue when a double or null object thinks it should respond to `__getobj__`. I'll need to think the implementation through a bit more to prevent that from being a problem. Closing for now.

myronmarston:
I think this would work:

``` ruby
if defined?(::Delegate) && ::Delegate === object
  # ...
else
  # ...
end
```

ahorner:
That approach does seem considerably safer. I've gone with that implementation.

ahorner:
There's one failure in the ree build which seems unrelated to my changes here.

lib/rspec/support/object_formatter.rb
```diff
@@ -88,6 +88,21 @@ def pretty_print(pp)
           pp.text inspection
         end
       end
+
+      # @private
+      DelegatingInspector = Struct.new(:object) do
+        def inspect
+          if defined?(::Delegator) && ::Delegator === object
+            "#<#{object.class}(#{DelegatingInspector.new(object.__getobj__).inspect})>"
```
myronmarston: Instead of just calling `inspect` on the wrapped object, it would be nice if the object was sent back through `prepare_for_inspection` above so that if you are wrapping one of the types of objects we have custom formatting for, that custom formatting will be displayed.  (And a spec for this case -- such as a delegator wrapping an array of time objects -- would be great, too).

lib/rspec/support/object_formatter.rb
```diff
@@ -88,6 +88,21 @@ def pretty_print(pp)
...
+          else
+            object.inspect
+          end
+        end
+
+        def pretty_print(pp)
+          pp.text inspect
+        end
```
myronmarston: Is there a spec that requires this method to pass?  I think you'd have to add a spec to the differ for it.  See here for an example:

https://github.com/rspec/rspec-support/blob/400b69dd70f1c1321f82ccabd882f144a610c1eb/spec/rspec/support/differ_spec.rb#L264-L320

spec/rspec/support/object_formatter_spec.rb
```diff
@@ -95,6 +95,18 @@ def with_date_loaded
...
+        let(:object) { Object.new }
+        let(:delegator) do
+          require "delegate"
+          SimpleDelegator.new(object)
+        end
+
+        it 'includes the delegator class in the description' do
+          expect(ObjectFormatter.format(delegator)).to eq "#<SimpleDelegator(#{object.inspect})>"
+        end
+      end
```
myronmarston: It would be nice to also have a spec for the case where `Delegator` is not defined, since the `defined?(::Delegator)` check is essential but not covered by any test.  You can simulate it not being defined by using `hide_const("Delegator")` in a spec.

lib/rspec/support/object_formatter.rb
ahorner: Done.

lib/rspec/support/object_formatter.rb
ahorner: Added.

spec/rspec/support/object_formatter_spec.rb
ahorner: And done.

myronmarston:
This LGTM.  Do you mind squashing it into one commit?

The travis failure is unrelated.  I'm going to work on a fix for that.

myronmarston:
There's a fix for the 1.8.7 failure in #216.  Hopefully we can merge that ASAP and then you can rebase against master.

myronmarston:
#216 is merged so you can rebase against master to get the fix and hopefully make your build go green.

ahorner:
I've rebased everything down into a single commit. Let me know if you need anything else.

PR #216, opened 2015-06-17
Fix spec so it always passes on 1.8.7.

myronmarston:
It failed on travis:

https://travis-ci.org/rspec/rspec-support/jobs/67086677

Hashes are unordered on 1.8.7 so the output order is non-deterministic.
For this spec it is not important that the hash has multiple entries,
so limiting it to one makes the output on all rubies deterministic.


myronmarston:
Ready for review, @rspec/rspec.  Would be nice to merge ASAP to unblock #215.

JonRowe:
I guess one key'd hashes are just as valid as multiple here :)

Issue #220, opened 2015-07-04
Support::OS.windows? is not returning correct information; should test for ! nil 


weedySeaDragon:
The Regexp pattern matching operator in <code>Support::OS.windows?</code> returns the position of the first match, so a result of 0 (zero) is valid -- it means that there *was* a match with the first character of a pattern in the RegExp.  If no match is found, the pattern match operator will return *nil.*

The method should instead test to see if the pattern matching does not return nil:
```
   def windows?
       !(RbConfig::CONFIG['host_os'] =~ /cygwin|mswin|mingw|bccwin|wince|emx/).nil?
   end
```

I'm running Windows 7. Using IRB, you can see that RbConfig correctly reports my host_os.  The pattern matcher correctly returns 0 -- which is the position of the first char of the string that matches "mingw32".   RSpec::Support::OS.windows? says that I'm not running windows.
The last line shows the change that should be made to correctly evaluate the result of the pattern matching and whether the host_os is Windows.
```
>> RbConfig::CONFIG['host_os'] 
=> "mingw32"
>> RbConfig::CONFIG['host_os'] =~ /cygwin|mswin|mingw|bccwin|wince|emx/
=> 0
>> RSpec::Support::OS.windows?
=> 0
>> RSpec::Support::OS.windows? == true
=> false
>> !(RbConfig::CONFIG['host_os'] =~ /cygwin|mswin|mingw|bccwin|wince|emx/).nil?
=> true
```

myronmarston:
It's working as designed.  `0` is a truthy value in ruby.  It's common in Ruby to just use a predicate's truthiness value in a conditional (e.g. `if RSpec::Support::OS.windows?`).  I searched our code and confirmed that that's the only way we are using it:


```
rspec-core/lib/rspec/core/configuration.rb
715:        if RSpec::Support::OS.windows? && !ENV['ANSICON']
1705:      if RSpec::Support::OS.windows?

rspec-core/lib/rspec/core/shell_escape.rb
12:      if RSpec::Support::OS.windows?

rspec-core/spec/integration/persistence_failures_spec.rb
57:        if RSpec::Support::OS.windows?

rspec-core/spec/rspec/core/backtrace_formatter_spec.rb
143:        it "excludes lines from rspec libs by default", :unless => RSpec::Support::OS.windows? do

rspec-core/spec/rspec/core/bisect/runner_spec.rb
240:      RSpec::Support::OS.windows? || RSpec::Support::Ruby.jruby?

[...]

PR #221, opened 2015-07-06
Make #windows? return true / false

JonRowe:
As suggested by @myronmarston in #220

### DIFF
--- a/lib/rspec/support/ruby_features.rb
+++ b/lib/rspec/support/ruby_features.rb
@@ -9,7 +9,7 @@ module RSpec
       module_function
 
       def windows?
-        RbConfig::CONFIG['host_os'] =~ /cygwin|mswin|mingw|bccwin|wince|emx/
+        !!(RbConfig::CONFIG['host_os'] =~ /cygwin|mswin|mingw|bccwin|wince|emx/)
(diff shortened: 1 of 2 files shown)


lib/rspec/support/ruby_features.rb
```diff
@@ -9,11 +9,11 @@ module OS
       module_function
 
       def windows?
-        RbConfig::CONFIG['host_os'] =~ /cygwin|mswin|mingw|bccwin|wince|emx/
+        !!(RbConfig::CONFIG['host_os'] =~ /cygwin|mswin|mingw|bccwin|wince|emx/)
       end
 
       def windows_file_path?
-        ::File::ALT_SEPARATOR == '\\'
+        !!(::File::ALT_SEPARATOR == '\\')
```
myronmarston: `String#==` already returns `true` or `false` so the `!!` is unnecessary here.

JonRowe: Derp, for some reason I read this as a regexp.

JonRowe: Updated

myronmarston:
LGTM.

PR #225, opened 2015-07-20
Work around JRuby `Method#parameters` bug.

myronmarston:
See jruby/jruby#2816 and rspec/rspec-mocks#919 for
a discussion of the bug. This is affecting us now that
I’m trying to enable partial double verification for
rspec-core (rspec/rspec-core#2032) so I thought it worth
addressing.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,11 @@ Enhancements:
 * Improve formatting of `Delegator` based objects (e.g. `SimpleDelgator`) in
   failure messages and diffs. (Andrew Horner, #215)
 
+Bug Fixes:
+
+* Work around bug in JRuby that reports that `attr_writer` methods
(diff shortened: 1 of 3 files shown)


JonRowe:
Seems to work on JRuby, just needs some rubocop violations tidied up :) (or turned off ^_^)

myronmarston:
I refactored the solution to no longer cause the rubocop violations and to use `super` instead of `alias`.

JonRowe:
Looks good to me but it's doing something strange on JRuby 1.8.7 still

myronmarston:
> Looks good to me but it's doing something strange on JRuby 1.8.7 still

Thanks, I added another bit to the conditional. Hopefully it'll work now.

JonRowe:
LGTM

PR #232, opened 2015-08-01
Extracts checks into common helper

AEgan:
rspec-rails can't run all library wide checks due to rails warnings and other dependencies.
This preps to allow for some specs to be extracted by rspec-rails until the aforementioned
is fixed


cupakromer:
Nice work :smile_cat:

PR #233, opened 2015-08-02
Update rubocop to latest with Ruby 2.2 support

cupakromer:
This updates Rubocop from 0.23.0 to 0.32.1 (most recent).

Unfortunately, in order to support this new version, which includes several bug fixes and support for Ruby 2.2, we need to drop support on Ruby 1.9.2. In fact Rubocop dropped 1.9.2 support as of version 0.25.0.

Some additional new cops are now available:
- Metrics/AbcSize (Assignment Branch Condition)
- Metrics/PerceivedComplexity

There are three additional new cops but they have been disabled (these settings should be ported to rspec-dev as part of the project defaults):
- Lint/NonLocalExitFromIterator
  
  This has been disabled due it's inability to distinguish between when / where `return` is sent from within a block. For example, a DSL block which has an explicit return in a method definition triggers this cop:
  
  ``` ruby
  some_method do
    def a_local_method
      return if true # triggers cop even though `return` will not exit from the block
    end
  end
  ```
- Style/ParallelAssignment
  
  Rubocop suggested this because it believed that parallel assignment was slower than sequential assignment. Unfortunately the benchmark they reference is flawed due to how Ruby handles implicit returns. It turns out that parallel assignment is _faster_.
- Style/StringLiteralsInInterpolation
  
  We previously disabled the related cop Style/StringLiterals. We simply don't care about single vs double qoutes.

See also: rspec/rspec-dev#134

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,8 @@ cache:
     - ../bundle
 before_install:
   - "script/clone_all_rspec_repos"
-  # Downgrade bundler to work around https://github.com/bundler/bundler/issues/3004
   # Note this doesn't work on JRUBY 2.0.0 mode so we don't do it
-  - if [ -z "$JRUBY_OPTS" ]; then gem install bundler -v=1.5.3 && alias bundle="bundle _1.5.3_"; fi
+  - if [ "jruby" != "$TRAVIS_RUBY_VERSION" ]; then gem install bundler; fi
 bundler_args: "--binstubs --standalone --without documentation --path ../bundle"
 script: "script/run_build"
 rvm:
--- a/Gemfile
+++ b/Gemfile
@@ -16,6 +16,11 @@ end
 ### dep for ci/coverage
 gem 'simplecov', '~> 0.8'
 
-gem 'rubocop', "~> 0.23.0", :platform => [:ruby_19, :ruby_20, :ruby_21]
+# There is no platform :ruby_193 and Rubocop only supports >= 1.9.3
+unless RUBY_VERSION == "1.9.2"
+  gem "rubocop",
+      "~> 0.32.1",
+      :platform => [:ruby_19, :ruby_20, :ruby_21, :ruby_22]
+end
 
 eval File.read('Gemfile-custom') if File.exist?('Gemfile-custom')
--- a/lib/rspec/support/ruby_features.rb
+++ b/lib/rspec/support/ruby_features.rb
@@ -6,7 +6,7 @@ module RSpec
     #
     # Provides query methods for different OS or OS features.
     module OS
-      module_function
+    module_function
 
       def windows?
         !!(RbConfig::CONFIG['host_os'] =~ /cygwin|mswin|mingw|bccwin|wince|emx/)
@@ -21,7 +21,7 @@ module RSpec
     #
     # Provides query methods for different rubies
     module Ruby
-      module_function
+    module_function
 
       def jruby?
         RUBY_PLATFORM == 'java'
@@ -45,7 +45,7 @@ module RSpec
     # Provides query methods for ruby features that differ among
     # implementations.
     module RubyFeatures
-      module_function
+    module_function
 
       def optional_and_splat_args_supported?
         Method.method_defined?(:parameters)
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -3,6 +3,14 @@ inherit_from: .rubocop_rspec_base.yml
 Lint/NonLocalExitFromIterator:
   Enabled: false
 
+# Over time we'd like to get this down, but this is what we're at now.
+Metrics/AbcSize:
+  Max: 22
+
+# Over time we'd like to get this down, but this is what we're at now.
+Metrics/PerceivedComplexity:
+  Max: 9
+
 # We don't care about single vs double qoutes.
 Style/StringLiteralsInInterpolation:
   Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,1 +1,4 @@
 inherit_from: .rubocop_rspec_base.yml
+
+Lint/NonLocalExitFromIterator:
+  Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,3 +2,7 @@ inherit_from: .rubocop_rspec_base.yml
 
 Lint/NonLocalExitFromIterator:
   Enabled: false
+
+# We don't care about single vs double qoutes.
+Style/StringLiteralsInInterpolation:
+  Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -14,3 +14,6 @@ Metrics/PerceivedComplexity:
 # We don't care about single vs double qoutes.
 Style/StringLiteralsInInterpolation:
   Enabled: false
+
+Style/SymbolProc:
+  Enabled: false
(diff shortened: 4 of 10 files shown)


cupakromer:
:frowning: way too many builds. A bit tired so was thrashing trying to get the travis config working. Now that things are green, I'm grabbing a few hours of sleep and will get this cleaned up in the AM.

Also, I'm temporarily dumping the parallel-vs-sequential assignment and symbol-to-proc-vs-block benchmarks to stdout in the `script/run_build`. You can see how all of the different Ruby versions shape up in the last travis build: https://travis-ci.org/rspec/rspec-support/builds/73746126

I skimmed the benchmarks and it looks like the general trends are:

- sequential assignment & blocks are faster on 1.8.7, REE, JRubys
- parallel assignment & symbol-to-proc are faster on 1.9.2 through 2.2

There were a lot of numbers to skim over tonight, but I think the benchmarks show that sequential vs parallel assignment really doesn't matter much for us. So we can disable that cop and use what seems best for the situation.

As for the block vs symbol-to-proc, we just need to decide if / when we are going to start following the optimizations for the newer Rubies. Thoughts?

/cc @rspec/rspec

cupakromer:
This is good to go.

.travis.yml
```diff
@@ -8,9 +8,8 @@ cache:
     - ../bundle
 before_install:
   - "script/clone_all_rspec_repos"
-  # Downgrade bundler to work around https://github.com/bundler/bundler/issues/3004
   # Note this doesn't work on JRUBY 2.0.0 mode so we don't do it
-  - if [ -z "$JRUBY_OPTS" ]; then gem install bundler -v=1.5.3 && alias bundle="bundle _1.5.3_"; fi
+  - if [ "jruby" != "$TRAVIS_RUBY_VERSION" ]; then gem install bundler; fi
```
JonRowe: This doesn't do the same thing, can we now safely use the later version of bundler on JRuby?

Gemfile
```diff
@@ -16,6 +16,11 @@ end
 ### dep for ci/coverage
 gem 'simplecov', '~> 0.8'
 
-gem 'rubocop', "~> 0.23.0", :platform => [:ruby_19, :ruby_20, :ruby_21]
+# There is no platform :ruby_193 and Rubocop only supports >= 1.9.3
+unless RUBY_VERSION == "1.9.2"
+  gem "rubocop",
+      "~> 0.32.1",
+      :platform => [:ruby_19, :ruby_20, :ruby_21, :ruby_22]
```
JonRowe: Surely all of this should be done from rspec-dev as it'll be the same across the gems

lib/rspec/support/ruby_features.rb
```diff
@@ -45,7 +45,7 @@ def mri?
     # Provides query methods for ruby features that differ among
     # implementations.
     module RubyFeatures
-      module_function
+    module_function
```
JonRowe: I personally prefer this indented, can we modify the rule instead? /cc @rspec/owners

JonRowe:
Good work on the benchmarks, but shouldn't we be doing the config via rspec-dev as its centralised? Also I disagree with one of the style changes but I'm willing to go with the majority obviously :)

.travis.yml
cupakromer: The comment above states that the older version of bundler is necessary due to an issue which is fixed. Though now I see that the default bundler, 1.7.4, doesn't contain the fix. So I can adjust this to support both.

Gemfile
cupakromer: rspec-dev does not control the Gemfiles. But yes, I am currently in the process of updating rspec-dev with these related changes. We need to add things to the repos directly for now to update rubocop. I plan on pushing a rspec-dev master change after they are all merged.

lib/rspec/support/ruby_features.rb
cupakromer: Can you elaborate? It's an access modified just like `public`, `private`, `protected`. It seems a little odd here but when there are multiple modifiers it seems strange to have this indented but the others out-dented.

cupakromer:
> but shouldn't we be doing the config via rspec-dev as its centralised?

Bumping rubocop will cause all the builds to fail. We've opened up PRs on all repos that have these fixes as a stop-gap so the new cops can be configured and fixes made. I am going to have a PR open for rspec-mocks tonight, and a PR for rspec-dev with the "master" changes; which I plan to push out after the PRs get merged in so that the build stays happy.

.travis.yml
cupakromer: Actually I'm not sure there is anything to change. The comment was stating that we weren't running the custom lower bundler version on JRuby, so it was supposed to be using the Travis default of 1.7.4. Am I missing something here?

Gemfile
cupakromer: I have an idea how I can have rspec-dev control this. I'll include it in the PR later.

JonRowe: BTW would this work:
```Ruby
gem 'rubocop', "~> 0.23.0", :platform => [:ruby_19]
gem "rubocop", "~> 0.32.1", :platform => [:ruby_20, :ruby_21, :ruby_22]
```

lib/rspec/support/ruby_features.rb
JonRowe: Everywhere else we would do `class << self` indented and the the methods again

JonRowe:
Yeah but we can open the PRs from rspec-dev with the bump and then fix them in those PRS

Gemfile
cupakromer: Unfortunately no.

```
[!] There was an error parsing `Gemfile`: You cannot specify the same gem twice with different version requirements.
You specified: rubocop (~> 0.23.0) and rubocop (~> 0.32.1). Bundler cannot continue.

 #  from /rspec-dev/repos/rspec-support/Gemfile:26
 #  -------------------------------------------
 #  gem "rubocop", "~> 0.23.0", :platform => [:ruby_19]
 >  gem "rubocop", "~> 0.32.1", :platform => [:ruby_20, :ruby_21, :ruby_22]
 #  
 #  -------------------------------------------
```

I did a brief test and some of the cops / settings are not backwards compatible. I'm not sure I see a good way to keep this running on 1.9.2 with an older Rubocop.

lib/rspec/support/ruby_features.rb
cupakromer: They don't have the same meaning. `module_function` makes the methods available both as a method on the module and adds the methods as `private` when `include` in another class or module.

Gemfile
JonRowe: Well in that case I think it's ok to drop 1.9.2 with rubocop, better that than missing 2.2.2 support (which has bugged me for a while TBH)

lib/rspec/support/ruby_features.rb
JonRowe: It does? TIL

cupakromer: Yep see [`Module#module_function`](http://ruby-doc.org/core-2.2.2/Module.html#method-i-module_function). It's pretty handy when you have a set of utility functions that you want to reference but do not want to increase an object's public API.

benoittgt:
@JonRowe do you think fix conflicts in a new PR (from this one) could be a good idea? I can work on it.

It seems lot's of things have change since. Maybe it can be close?

JonRowe:
@benoittgt feel free to take a shot at rebasing this

benoittgt:
Closing for now. In favor of https://github.com/rspec/rspec-support/pull/350

PR #234, opened 2015-08-05
Allow warning behaviour to be configurable

JonRowe:
This emulates the failure notifier stuff for warnings to support rspec/rspec-core#2052

### DIFF
--- a/lib/rspec/support.rb
+++ b/lib/rspec/support.rb
@@ -78,6 +78,7 @@ module RSpec
       Thread.current[:__rspec] ||= {}
     end
 
+    # @api private
     def self.failure_notifier=(callable)
       thread_local_data[:failure_notifier] = callable
     end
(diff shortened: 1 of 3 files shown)


spec/rspec/support_spec.rb
```diff
@@ -180,5 +180,44 @@ def raise_instead_of_appending_to_failures_array
...
+
+      def warn_instead_of_appending_to_array
+        expect(::Kernel).to receive(:warn).with(a_string_including('some warning'))
+        avoid_changing { warnings }
+      end
+
+      it "defaults to warning with the provided text" do
+        expect(::Kernel).to receive(:warn).with(a_string_including('some warning'))
+        warning('some warning')
+      end
```
myronmarston: I think I'd rather the `output().to_stderr` matcher was use rather than mocking `::Kernel.warn`.  Mocking seems unnecessary here.

lib/rspec/support.rb
```diff
@@ -101,6 +101,17 @@ def self.with_failure_notifier(callable)
       self.failure_notifier = orig_notifier
     end
 
+    def self.warning_notifier=(callable)
+      thread_local_data[:warning_notifier] = callable
```
myronmarston: I'm not sure that we want to scope it to a thread like we did with the failure notifier.  After all, the usage of this in rspec-core is to have a global config that's intended to be set once during config time and then be left for the rest of the process.  It would be confusing for users to configure that and then a warning emitted from another thread and have it be printed rather than raise an error like they have configured.

For the failure notifier we needed it to be scoped to a thread because we mutate it for one example and then mutate it back, and the notifier tracks failures so there's all this mutable state being dealt with.

JonRowe:
Updated to address your feedback @myronmarston

JonRowe:
Anyone have time to review this @rspec/rspec

xaviershay:
Needs docs if it's going to be public API, LGTM otherwise.

PR #240, opened 2015-09-28
Expose wether exception cause is supported

JonRowe:
No changelog because this is an internal feature

### DIFF
--- a/lib/rspec/support/ruby_features.rb
+++ b/lib/rspec/support/ruby_features.rb
@@ -55,6 +55,16 @@ module RSpec
         respond_to?(:caller_locations, true)
       end
 
+      if Exception.method_defined?(:cause)
+        def supports_exception_cause?
+          true
+        end
+      else
+        def supports_exception_cause?
+          false
+        end
+      end
+
       if Ruby.mri?
         def kw_args_supported?
           RUBY_VERSION >= '2.0.0'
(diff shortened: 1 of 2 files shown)


JonRowe:
@samphippen care to give this a once over?

lib/rspec/support/ruby_features.rb
```diff
@@ -55,6 +55,18 @@ def caller_locations_supported?
         respond_to?(:caller_locations, true)
       end
 
+      begin
```
samphippen: What do you think about doing a `respond_to?` check here? I think that'd be a lot cleaner.

JonRowe: I preferred not instantiating an exception...

samphippen: oh, right, `Exception.instance_methods.include?(:cause)`?

myronmarston: `Exception.method_defined?(:cause)`

No need for a linear `include?` operation.

samphippen: :+1:

JonRowe: Done.

JonRowe:
Hmm randomly this is failing on windows but it's for an unrelated encoding spec..

myronmarston:
LGTM.  Merged.

PR #243, opened 2015-10-06
fix rack-cache on < 2.0.0

JonRowe:


myronmarston:
What does rack cache have to do with rspec-support?

JonRowe:
> Gem::InstallError: rack-cache requires Ruby version >= 2.0.0.
Using rspec-support 3.4.0.pre from source at /home/travis/build/rspec/rspec-support
An error occurred while installing rack-cache (1.3.0), and Bundler cannot
continue.
Make sure that `gem install rack-cache -v '1.3.0'` succeeds before bundling.

JonRowe:
I think its a dependency of a dependency

myronmarston:
Can you add a comment to the README explaining it's not a direct dependency?  It would be nice to note what it is a dependency of as well.

myronmarston:
Actually, my rspec-support `Gemfile.lock` doesn't have `rack-cache` at all so I really don't think that's needed here.  This might be needed in rspec-rails, though?

JonRowe:
That wouldn't explain mocks failure based on this though?

JonRowe:
Ah no you're right, it's just bundler mangling the output

PR #245, opened 2015-10-10
Add RubyFeatures.ripper_supported?

yujinakayama:
For https://github.com/rspec/rspec-core/pull/2083

### DIFF
--- a/lib/rspec/support/comparable_version.rb
+++ b/lib/rspec/support/comparable_version.rb
@@ -1,0 +1,46 @@
+module RSpec
+  module Support
+    # @private
+    class ComparableVersion
+      include Comparable
+
+      attr_reader :string
+
+      def initialize(string)
+        @string = string
+      end
+
+      def <=>(other)
+        other = self.class.new(other) unless other.is_a?(self.class)
+
+        return 0 if string == other.string
+
+        longer_segment_count = [self, other].map { |version| version.segments.count }.max
+
+        longer_segment_count.times do |index|
+          self_segment = segments[index] || 0
+          other_segment = other.segments[index] || 0
+
+          if self_segment.class == other_segment.class
+            result = self_segment <=> other_segment
+            return result unless result == 0
+          else
+            return self_segment.is_a?(String) ? -1 : 1
+          end
+        end
+
+        0
+      end
+
+      def segments
+        @segments ||= string.scan(/[a-z]+|\d+/i).map do |segment|
+          if segment =~ /\A\d+\z/
+            segment.to_i
+          else
+            segment
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/rspec/support/ruby_features.rb
+++ b/lib/rspec/support/ruby_features.rb
@@ -1,4 +1,5 @@
 require 'rbconfig'
+RSpec::Support.require_rspec_support "comparable_version"
 
 module RSpec
   module Support
@@ -61,6 +62,24 @@ module RSpec
         end
       else
         def supports_exception_cause?
+          false
+        end
+      end
+
+      ripper_requirements = [ComparableVersion.new(RUBY_VERSION) >= '1.9.2']
+
+      if Ruby.jruby?
+        ripper_requirements.push(ComparableVersion.new(JRUBY_VERSION) >= '1.7.5')
+        # Ripper on JRuby 9.0.0.0.rc1 or later reports wrong line number.
+        ripper_requirements.push(ComparableVersion.new(JRUBY_VERSION) < '9.0.0.0.rc1')
+      end
+
+      if ripper_requirements.all?
+        def ripper_supported?
+          true
+        end
+      else
+        def ripper_supported?
           false
         end
       end
--- a/spec/rspec/support/ruby_features_spec.rb
+++ b/spec/rspec/support/ruby_features_spec.rb
@@ -86,6 +86,59 @@ module RSpec
         it 'does not load Ripper' do
           expect { RubyFeatures.ripper_supported? }.not_to change { defined?(::Ripper) }
         end
+
+        describe 'Ripper' do
+          let(:line_number) do
+            token = tokens.first
+            location = token.first
+            location.first
+          end
+
+          let(:tokens) do
+            require 'ripper'
+            ::Ripper.lex('foo')
+          end
+
+          if Ruby.mri?
+            context 'on MRI' do
+              context '1.8.x', :if => RUBY_VERSION.start_with?('1.8.') do
+                it 'is not supported' do
+                  expect { tokens }.to raise_error(LoadError)
+                end
+              end
+
+              context '1.9.x or later', :if => RUBY_VERSION >= '1.9' do
+                it 'is supported' do
+                  expect(line_number).to eq(1)
+                end
+              end
+            end
+          end
+
+          if Ruby.jruby?
+            context 'on JRuby' do
+              context '1.7.x', :if => JRUBY_VERSION.start_with?('1.7.') do
+                context 'in 1.8 mode', :if => RUBY_VERSION.start_with?('1.8.') do
+                  it 'is not supported' do
+                    expect { tokens }.to raise_error(NameError)
+                  end
+                end
+
+                context 'in non 1.8 mode', :unless => RUBY_VERSION.start_with?('1.8.') do
+                  it 'is supported' do
+                    expect(line_number).to eq(1)
+                  end
+                end
+              end
+
+              context '9.x.x.x', :if => JRUBY_VERSION.start_with?('9.') do
+                it 'reports wrong line number' do
+                  expect(line_number).to eq(2)
+                end
+              end
+            end
+          end
+        end
       end
     end
   end
--- a/spec/rspec/support/ruby_features_spec.rb
+++ b/spec/rspec/support/ruby_features_spec.rb
@@ -81,6 +81,12 @@ module RSpec
           expect(RubyFeatures.caller_locations_supported?).to eq(RUBY_VERSION >= '2.0.0')
         end
       end
+
+      describe "#ripper_supported?" do
+        it 'does not load Ripper' do
+          expect { RubyFeatures.ripper_supported? }.not_to change { defined?(::Ripper) }
+        end
+      end
     end
   end
 end
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,7 @@ Enhancements:
 
 * Improve formatting of `Delegator` based objects (e.g. `SimpleDelgator`) in
   failure messages and diffs. (Andrew Horner, #215)
+* Add `ComparableVersion`. (Yuji Nakayama, #245)
 
 Bug Fixes:
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,7 @@ Enhancements:
 * Improve formatting of `Delegator` based objects (e.g. `SimpleDelgator`) in
   failure messages and diffs. (Andrew Horner, #215)
 * Add `ComparableVersion`. (Yuji Nakayama, #245)
+* Add `Ripper` support detection. (Yuji Nakayama, #245)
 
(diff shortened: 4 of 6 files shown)


lib/rspec/support/ruby_features.rb
```diff
@@ -65,6 +65,12 @@ def supports_exception_cause?
         end
       end
 
+      def ripper_supported?
+        # JRuby supports Ripper but it seems to be unstable
+        # and also reports wrong line number on JRuby 9.0.0.0.
+        defined?(RUBY_ENGINE) && RUBY_ENGINE == 'ruby' && RUBY_VERSION >= '1.9.2'
```
myronmarston: This is too bad about JRuby's ripper support not being good enough for us to use.  Has the issue been reported upstream?  I imagine @headius and co would want to know and do something about this.  Unfortunately, as this is currently written it'll lock JRuby users out from enjoying the benefits of the feature if/when JRuby's ripper is fixed.

Would it be possible to write some logic that detects if ripper is working on JRuby so that we allow it to be used if a user is running a future version of JRuby where the ripper issues have been fixed?  Although, the obvious ways to do that would involve loading ripper and testing it out, which I don't think we want to do...

spec/rspec/support/ruby_features_spec.rb
```diff
@@ -81,6 +81,12 @@ module Support
           expect(RubyFeatures.caller_locations_supported?).to eq(RUBY_VERSION >= '2.0.0')
         end
       end
+
+      specify "#ripper_supported? exists" do
+        RubyFeatures.ripper_supported?
+        # Ensure .ripper_supported? does not load Ripper.
+        expect { ::Ripper }.to raise_error(NameError)
```
myronmarston: Given that we now use ripper as part of the failure output, this expectation seems brittle and likely to fail if any earlier specs have failed and you're running the suite with a formatter that prints failures as they happen.  Instead, WDYT of this?

``` ruby
expect {
  RubyFeatures.ripper_supported?
}.not_to change { defined?(Ripper) }
```

The important thing is that `ripper_supported?` does not load it -- but if something else has already loaded it, that's not our concern here.

yujinakayama: :+1:

lib/rspec/support/ruby_features.rb
yujinakayama: Just reported the wrong line number issue https://github.com/jruby/jruby/issues/3386.

yujinakayama: > Would it be possible to write some logic that detects if ripper is working on JRuby so that we allow it to be used if a user is running a future version of JRuby where the ripper issues have been fixed? Although, the obvious ways to do that would involve loading ripper and testing it out, which I don't think we want to do...

I added examples to check if the Ripper behavior in rspec-support's spec rather than user's environement so that we can immediately notice the change on Travis CI in the future. By the way I noticed that we do not have CI for JRuby 9.0.0.0.

yujinakayama: Also, I confirmed specs in https://github.com/rspec/rspec-core/pull/2083 properly passed on JRuby 1.7.14 - 1.7.22.

lib/rspec/support/ruby_features.rb
```diff
@@ -65,6 +65,25 @@ def supports_exception_cause?
...
       end
 
+      if Ruby.mri? && RUBY_VERSION >= '1.9.0'
+        def ripper_supported?
+          true
+        end
+      elsif Ruby.jruby? &&
+            defined?(Gem::Version) &&
+            Gem::Version.new(JRUBY_VERSION) >= Gem::Version.new('1.7.14') &&
+            !JRUBY_VERSION.start_with?('9.')
```
myronmarston: This conditional will return `false` if rubygems is not loaded.  It should not be a requirement that rubygems is loaded to use this feature.  Users may want to run `rspec` w/ the `--disable=gems` flag and that should work, IMO.  I think you can use our version checker instead:

https://github.com/rspec/rspec-support/blob/master/lib/rspec/support/version_checker.rb

lib/rspec/support/ruby_features.rb
```diff
@@ -65,6 +65,25 @@ def supports_exception_cause?
...
+        # JRuby 1.7.5 - 1.7.13 actually supports Ripper but it's too slow.
```
myronmarston: How slow?

JonRowe: Benchmarked above says 1/10 of the speed of normal Ruby, but could this be a compile time issue?

yujinakayama: From my understanding Ripper on JRuby performs parsing purely in runtime so I don't think it's a compile time issue.

yujinakayama:
Not sure why but the CI is failing now, even on master branch...

```
Running specs for rspec-core
~/build/rspec/rspec-core ~/build/rspec/rspec-support
`ruby_22` is not a valid platform. The available options are: [:ruby, :ruby_18,
:ruby_19, :ruby_20, :ruby_21, :mri, :mri_18, :mri_19, :mri_20, :mri_21, :rbx,
:jruby, :jruby_18, :jruby_19, :mswin, :mingw, :mingw_18, :mingw_19, :mingw_20,
:mingw_21, :x64_mingw, :x64_mingw_20, :x64_mingw_21]
The command "script/run_build" exited with 4.
cache.2
```

lib/rspec/support/ruby_features.rb
yujinakayama: :+1:

myronmarston:
Those are bundler platform values. Maybe a bundler regression? Did the bundler version change?

yujinakayama:
> Did the bundler version change?

No.

Passed build: https://travis-ci.org/rspec/rspec-support/jobs/83149928

```
$ ruby --version
ruby 2.2.0p0 (2014-12-25 revision 49005) [x86_64-linux]
$ rvm --version
rvm 1.26.10 (latest-minor) by Wayne E. Seguin <wayneeseguin@gmail.com>, Michal Papis <mpapis@gmail.com> [https://rvm.io/]
$ bundle --version
Bundler version 1.7.9
$ gem --version
2.4.5
```

Failed build: https://travis-ci.org/rspec/rspec-support/jobs/83956734

```
$ ruby --version
ruby 2.2.0p0 (2014-12-25 revision 49005) [x86_64-linux]
$ rvm --version
rvm 1.26.10 (latest-minor) by Wayne E. Seguin <wayneeseguin@gmail.com>, Michal Papis <mpapis@gmail.com> [https://rvm.io/]
$ bundle --version
Bundler version 1.7.9
$ gem --version
2.4.5
```

lib/rspec/support/ruby_features.rb
JonRowe: I mean that JRuby usually has a higher startup cost and I wondered if that was skewing the benchmark.

myronmarston:
Looking at the benchmark, I can see that ripper is certainly much slower on those JRuby 1.7.x releases but I'm not sure that it's so slow that it makes sense to disable the feature entirely.  Consider that the benchmark shows 240-250 ripper parsing operations per second on those versions of JRuby.  That's only  4 ms or so.  I think the productivity boost from having the full code snippet is going to save the user much more than 4 ms of time.  Besides, in the rspec-core PR you've implemented a source cache so the slow parsing cost will only be paid once per distinct file that has a failure.  In my experience, it's a lot more common to have a few files, each with many failures, than to have many files, each with few failures.  This slowness is also not a general slowdown we're adding for all `rspec` runs -- it just comes in to play when there's a failure and we're trying to provide maximally useful failure output.

lib/rspec/support/ruby_features.rb
```diff
@@ -65,6 +66,29 @@ def supports_exception_cause?
...
+          VersionChecker.new('JRuby', JRUBY_VERSION, '1.7.14').check_version!
+
+          # Ripper on JRuby 9.0.0.0 reports wrong line number.
+          raise UnsupportedVersionError if JRUBY_VERSION.start_with?('9.')
+        end
+
+        def ripper_supported?
+          true
+        end
+      rescue LibraryVersionTooLowError, UnsupportedVersionError
```
myronmarston: Exceptions for control flow is always one of those things that feels smelly.  Instead, WDYT of making `VersionChecker#compare_version` public and using that to find out how the version compares w/o needing to use exceptions?

Actually, it looks like nothing is using the version checker so you should feel free to change it's API to whatever you need :).

lib/rspec/support/ruby_features.rb
JonRowe: I mean theres usually a period where JRuby is slow after startup but optimises after...

yujinakayama:
* Replaced `VersionChecker` with the new `ComparableVersion`.
* Changed the Ripper support on JRuby to `>= 1.7.5 && < 9.0.0.0`.

lib/rspec/support/comparable_version.rb
```diff
@@ -0,0 +1,35 @@
...
+
+        0
+      end
+
+      def segments
+        @segments ||= string.split('.').map { |v| v.to_i }
+      end
+    end
+  end
+end
```
myronmarston: This class looks nice and simple, but will it handle version numbers with `.pre`, `.rc1`, `.beta`, etc in the name?  The `segments` method uses `to_i` which will silently convert those to `0`.  Looking at the [JRuby downloads page](http://jruby.org/files/downloads/index.html), I see version numbers like `9.0.0.0.pre1` and `9.0.0.0.rc1`.  Since we pass the `JRUBY_VERSION` to this I think we need to handle those versions properly.

lib/rspec/support/ruby_features.rb
```diff
@@ -65,6 +66,24 @@ def supports_exception_cause?
...
+
+      if ripper_requirements.all?
+        def ripper_supported?
+          true
+        end
+      else
+        def ripper_supported?
+          false
+        end
+      end
```
myronmarston: Nice!  The `ripper_requirements.all?` thing works really simply.  I like it :).

myronmarston:
This LGTM besides the concern about pre-release words in version numbers.

lib/rspec/support/version_checker.rb
```diff
@@ -1,53 +0,0 @@
-module RSpec
-  module Support
-    LibraryVersionTooLowError = Class.new(StandardError)
-
-    # @private
-    class VersionChecker
```
JonRowe: I think this is used outside of `rspec-support` (I'm thinking `rspec-rails` at least) so we might need to leave this in until we've switched those instances over.

yujinakayama: The current `rspec-rails` is not using `VersionChecker` and the `rspec-support` dependency is [pinned to major and minor versions](https://github.com/rspec/rspec-rails/blob/078b29fefb129b7942ba512cc0173b0c6d20c582/rspec-rails.gemspec#L35), so I think removing `VersionChecker` does not break anything.

lib/rspec/support/comparable_version.rb
yujinakayama: Done.

myronmarston:
This LGTM apart from the build failure.

lib/rspec/support/version_checker.rb
yujinakayama: I was wrong. It's still used [here](https://github.com/rspec/rspec-rails/blob/078b29fefb129b7942ba512cc0173b0c6d20c582/lib/rspec/rails/vendor/capybara.rb#L12-L13). The `vendor` directory was ignored by my `ag`...

JonRowe: I thought it was used somewhere! It was created for `rspec-rails` originally :P

JonRowe:
Want to rebase this off master and see if it works now?

yujinakayama:
Done

JonRowe:
Sweet as, :green_apple:

PR #248, opened 2015-10-20
FuzzyMatcher is friendly to custom classes subclassed from Array

alexdowad:
The semantics of `FuzzyMatcher`'s comparison algorithm are problematic for custom classes which subclass from `Array`. This commit adjusts `FuzzyMatcher` to make its behavior on subclasses of `Array` more intuitive.

This is the commit log message:

When comparing subclasses of `Array`, `FuzzyMatcher` would call `#to_a` on one object
but not both, which could cause the comparison to fail if `#to_a` was overridden.
Additionally, some subclasses of `Array` may define their own idea of 'equality',
which should be respected by `FuzzyMatcher`.

On the other hand, if the 'expected' value is an `Array`, it is reasonable to
coerce the subclass of `Array` to an actual `Array`. Since `Enumerable` does not
define `#to_ary`, but only `#to_a`, use `#to_a` for this purpose.

### DIFF
--- a/lib/rspec/support/fuzzy_matcher.rb
+++ b/lib/rspec/support/fuzzy_matcher.rb
@@ -6,9 +6,11 @@ module RSpec
     module FuzzyMatcher
       # @api private
       def self.values_match?(expected, actual)
-        if Hash === actual
-          return hashes_match?(expected, actual) if Hash === expected
-        elsif Array === expected && Enumerable === actual && !(Struct === actual)
+        if actual.instance_of?(Hash)
+          return hashes_match?(expected, actual) if expected.instance_of?(Hash)
+        elsif expected.instance_of?(Array) && actual.instance_of?(Array)
+          return arrays_match?(expected, actual)
+        elsif expected.instance_of?(Array) && Enumerable === actual && !(Struct === actual)
           return arrays_match?(expected, actual.to_a)
         end
 
--- a/spec/rspec/support/fuzzy_matcher_spec.rb
+++ b/spec/rspec/support/fuzzy_matcher_spec.rb
@@ -145,6 +145,64 @@ module RSpec
         end
       end
 
+      context "when given objects which subclass from Array" do
+        context "which override #to_a and #to_ary" do
+          let(:my_array) do
+            Class.new(Array) do
+              # when both objects are subclassed from Array,
+              # the comparison should not try to coerce one or both to Array
+              def to_a
+                raise "No you don't"
+              end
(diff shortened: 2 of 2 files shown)


lib/rspec/support/fuzzy_matcher.rb
```diff
@@ -6,9 +6,11 @@ module Support
...
       # @api private
       def self.values_match?(expected, actual)
-        if Hash === actual
-          return hashes_match?(expected, actual) if Hash === expected
-        elsif Array === expected && Enumerable === actual && !(Struct === actual)
+        if actual.instance_of?(Hash)
+          return hashes_match?(expected, actual) if expected.instance_of?(Hash)
+        elsif expected.instance_of?(Array) && actual.instance_of?(Array)
+          return arrays_match?(expected, actual)
+        elsif expected.instance_of?(Array) && Enumerable === actual && !(Struct === actual)
```
myronmarston: We avoid `object.instance_of?(klass)` checks, opting for `klass === object` instead.  These should generally be equivalent, but in the case of `object.instance_of?`, since `object` could be anything, we don't know if it will even respond to `instance_of?` or if it's been redefined to do something different.  (Users pass all manner of objects to RSpec...).  For example, `BasicObject === BasicObject.new` returns `true` while `BasicObject.new.instance_of?(BasicObject)` raises a `NoMethodError`.

Please switch all these `instance_of?(klass)` checks to use `klass ===` instead.

alexdowad: OK. The problem with `Class#===` is that it returns `true` not just for instances of `klass`, but for instances of *subclasses* of `klass`. The whole point of this PR is that `FuzzyMatcher`'s matching doesn't work well with subclasses of `Array` which may have redefined `#to_a` or `#==`.

This is a tricky situation! We're working in a very dynamic language, where any method on any object can be redefined to do anything at all. I see why you like `Class#===`. Because it is defined on `Class`, and not the object being tested, you don't need to make any assumptions at all about what crazy things might have been done to the object under test.

An alternative which works in Ruby 2.0 and up is `Kernel.instance_method(:instance_of?).bind(obj).call(Array)`.

Another alternative is to use an interpreter-specific extension module which reaches "under the hood" to pick out the actual class of an object. With this technique, you can write code which works even with `BasicObject` instances, or other classes which have had the built-in `Object` methods redefined to do crazy things.

Any thoughts?

alexdowad:
Closing due to lack of activity.

PR #250, opened 2015-11-04
Implement granular expectations for method signatures

sleepingkingstudios:
Supports https://github.com/rspec/rspec-expectations/issues/846.

Allows `MethodSignatureVerifier` to verify:
- Heterogenous min and max arguments counts (e.g. functions with optional arguments).
- Optional and/or required keyword arguments if supported by Ruby runtime.
- Splatted arguments and/or keyword arguments.

For an example of use, see https://github.com/rspec/rspec-expectations/compare/master...sleepingkingstudios:respond_to-keyword-arguments?expand=1

### DIFF
--- a/lib/rspec/support/method_signature_verifier.rb
+++ b/lib/rspec/support/method_signature_verifier.rb
@@ -27,6 +27,8 @@ module RSpec
       end
 
       def valid_non_kw_args?(positional_arg_count)
+        return true if positional_arg_count.nil?
+
         min_non_kw_args <= positional_arg_count &&
           positional_arg_count <= max_non_kw_args
       end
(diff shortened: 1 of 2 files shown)


lib/rspec/support/method_signature_verifier.rb
```diff
@@ -26,9 +26,13 @@ def non_kw_args_arity_description
         end
       end
 
-      def valid_non_kw_args?(positional_arg_count)
+      def valid_non_kw_args?(positional_arg_count, optional_max_arg_count=nil)
```
JonRowe: You should be able to set the default for `optional_max_arg_count` to `positional_arg_count` which should eliminate the `||=` below.

lib/rspec/support/method_signature_verifier.rb
```diff
@@ -26,9 +26,13 @@ def non_kw_args_arity_description
...
+        return true if positional_arg_count.nil?
+
+        optional_max_arg_count ||= positional_arg_count
```
JonRowe: Can you either collapse this below or use the suggestion for default arg rather than shadowing variables.

JonRowe:
thanks! merging despite build failure as it's a known bundler issue

Issue #252, opened 2015-12-05
Ensure output of ObjectFormatter.format has reasonable output length


samphippen:
[This method](https://github.com/rspec/rspec-support/blob/master/lib/rspec/support/object_formatter.rb#L7) needs some kind of length capping that ensures it's human-readable.

See: https://github.com/rspec/rspec-rails/issues/1490 for en example of the problem.

samphippen:
For anyone who might be new, looking to help with this issue. You can find our contributing doc here:
https://github.com/rspec/rspec-support/blob/master/CONTRIBUTING.md.

I think there are a few tasks here:

- [ ] Add an `attr_accessor :max_formatted_output_length` which is used to range limit the output string from the inspect call
- [ ] Make the output length not cap if that attribute is set to nil
- [ ] expose an API for this in RSpec Expectations configuration
- [ ] set some reasonable default in RSpec Expectations configuration

samphippen:
We probably only need to do those first two here, and someone else can do the stuff on RSpec Expectations.

liam-m:
I think I'd be able to do the first two.

I've cloned the repo and run bundle install.
I then ran the tests (using ```rake```) but had a failure:

    Run options: include {:focus=>true}

    All examples were filtered out; ignoring {:focus=>true}

    Randomized with seed 18660
    ..............................................................................................................................................................................................................................................F.................................................*.............................

    Pending: (Failures listed here are expected and do not affect your suite's status)

      1) RSpec::Support::ShellOut can shell out to ruby with the current load path
         # Need to investigate why this is failing -- see https://travis-ci.org/rspec/rspec-core/jobs/60327106 and https://travis-ci.org/rspec/rspec-support/jobs/60296920 for examples
         # ./spec/rspec/support/spec/shell_out_spec.rb:22


    Failures:

      1) RSpec::Support::Differ#diff copes with encoded strings
         Failure/Error: DEFAULT_FAILURE_NOTIFIER = lambda { |failure, _opts| raise failure }
     
           expected
           "" (UTF-8) to be identical to
           "\n@@ -1,2 +1,2 @@\n-Tu avec carte {count} item has\n+Tu avec cart\u00E9 {count} it\u00E9m has\n" (UTF-16LE)
           The exact bytes are printed below for more detail:
           []
[...]

PR #256, opened 2015-12-11
Truncate ObjectFormatter.format's output (#252)

liam-m:
Initial changes and a simple test for
https://github.com/rspec/rspec-support/issues/252

### DIFF
--- a/lib/rspec/support/object_formatter.rb
+++ b/lib/rspec/support/object_formatter.rb
@@ -2,15 +2,41 @@ module RSpec
   module Support
     # Provide additional output details beyond what `inspect` provides when
     # printing Time, DateTime, or BigDecimal
-    module ObjectFormatter
-      # @api private
+    # @api private
+    class ObjectFormatter
+      attr_accessor :max_formatted_output_length
+
+      def initialize(max_formatted_output_length=200)
+        @max_formatted_output_length = max_formatted_output_length
+      end
+
+      # Methods are deferred to a default instance of the class to maintain the interface
+      # For example, calling ObjectFormatter.format is still possible
+      @default_instance = new(200)
+
+      ELLIPSIS = "..."
+
+      def format(object)
+        if max_formatted_output_length.nil?
+          return prepare_for_inspection(object).inspect
+        else
+          formatted_object = prepare_for_inspection(object).inspect
+          if formatted_object.length < max_formatted_output_length
+            return formatted_object
+          else
+            beginning = formatted_object[0 .. max_formatted_output_length / 2]
+            ending = formatted_object[-max_formatted_output_length / 2 ..-1]
+            return beginning + ELLIPSIS + ending
+          end
+        end
+      end
+
       def self.format(object)
-        prepare_for_inspection(object).inspect
+        @default_instance.format(object)
       end
 
       # rubocop:disable MethodLength
 
-      # @private
       # Prepares the provided object to be formatted by wrapping it as needed
       # in something that, when `inspect` is called on it, will produce the
       # desired output.
@@ -20,7 +46,7 @@ module RSpec
       # with custom items that have `inspect` defined to return the desired output
       # for that item. Then we can just use `Array#inspect` or `Hash#inspect` to
       # format the entire thing.
-      def self.prepare_for_inspection(object)
+      def prepare_for_inspection(object)
         case object
         when Array
           return object.map { |o| prepare_for_inspection(o) }
@@ -44,33 +70,41 @@ module RSpec
       end
       # rubocop:enable MethodLength
 
-      # @private
-      def self.prepare_hash(input)
+      def self.prepare_for_inspection(object)
+        @default_instance.prepare_for_inspection(object)
+      end
+
+      def prepare_hash(input)
         input.inject({}) do |hash, (k, v)|
           hash[prepare_for_inspection(k)] = prepare_for_inspection(v)
           hash
         end
       end
 
+      def self.prepare_hash(input)
+        @default_instance.prepare_hash(input)
+      end
+
       TIME_FORMAT = "%Y-%m-%d %H:%M:%S"
 
       if Time.method_defined?(:nsec)
-        # @private
-        def self.format_time(time)
+        def format_time(time)
           time.strftime("#{TIME_FORMAT}.#{"%09d" % time.nsec} %z")
         end
       else # for 1.8.7
-        # @private
-        def self.format_time(time)
+        def format_time(time)
           time.strftime("#{TIME_FORMAT}.#{"%06d" % time.usec} %z")
         end
+      end
+
+      def self.format_time(time)
+        @default_instance.format_time(time)
       end
 
       DATE_TIME_FORMAT = "%a, %d %b %Y %H:%M:%S.%N %z"
       # ActiveSupport sometimes overrides inspect. If `ActiveSupport` is
       # defined use a custom format string that includes more time precision.
-      # @private
-      def self.format_date_time(date_time)
+      def format_date_time(date_time)
         if defined?(ActiveSupport)
           date_time.strftime(DATE_TIME_FORMAT)
         else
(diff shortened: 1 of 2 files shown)


lib/rspec/support/object_formatter.rb
```diff
@@ -3,9 +3,23 @@ module Support
     # Provide additional output details beyond what `inspect` provides when
     # printing Time, DateTime, or BigDecimal
     module ObjectFormatter
+      class << self
+        attr_accessor :max_formatted_output_length
+      end
```
myronmarston: Adding mutable state at the module level is always a bit of a code smell, and requires gymnastics such a the around hook in your test to isolate it.  Up to know the object formatter didn't have any state which is why it is just available as methods directly off of the module.  I think that instead of adding this module-level state, I'd advocate:

* Turning it into a class, that receives `max_formatted_output_length` as an initialization argument.
* The existing interface be preserved (as a facade of sorts) by keeping class methods that delegate to a default shared instance of the class.
* Then in tests it can just instantiate a different instance with a different `max_formatted_output_length` value, with no need to mutate it and reset it.

liam-m: Thank you for the recommendations, I agree it would be nicer to avoid adding state to the module.

* Should the max_formatted_ouput_length have a default value? If so, what should it be?
* Where should the shared instance be defined?
* I've used Forwardable for delegation but not to an instance - would it be possible to use Forwardable for that?

myronmarston: > Should the max_formatted_ouput_length have a default value?

Yes.

> If so, what should it be?

Off the top of my head, 200 characters sounds about right, but I'd have to see comparisons of output at different levels to say for sure.  Can we go with 200 for now, and then maybe after you have it working, you can paste snippets  or screenshots of example output at different length values?

> Where should the shared instance be defined?

Here is fine.  Something like:

``` ruby
@default_instance = new

def self.format(object)
  @default_instance.format(object)
end
```

> I've used Forwardable for delegation but not to an instance - would it be possible to use Forwardable for that?

I believe it would be possible, but we have a policy of using as little of the stdlib as possible.  If we used forwardable here, that would allow a gem author to write a gem that uses forwardable and forget to require it.  There RSpec test suite would still be green (since RSpec would load forwardable due to your use of it here) but their gem would be broken -- users of their gem that were not already requiring forwardable would get errors.  We try to avoid these situations by minimizing our use of the standard library.

lib/rspec/support/object_formatter.rb
```diff
@@ -3,9 +3,25 @@ module Support
...
       def self.format(object)
-        prepare_for_inspection(object).inspect
+        if max_formatted_output_length.nil?
+          return prepare_for_inspection(object).inspect
+        else
+          formatted_object = prepare_for_inspection(object).inspect
+          if formatted_object.length < max_formatted_output_length
+            return formatted_object
+          else
+            ellipsis = "\u2026".force_encoding('utf-8')
```
myronmarston: This could be a constant.

samphippen:
@liam-m 1.8.7 has no `force_encoding` method. What do you think about using the literal "..." instead on 1.8.7

JonRowe:
I think we should use a literal ... anyway, why wouldn't we?

samphippen:
@JonRowe I think the unicode ellipses is nice UX. Is this ready to merge?

JonRowe:
I meant use a literal Unicode ellipses char, the file is Utd-8 so no need
for the encoding right? Especially as the output encoding isn't going to
always be Utf-8
>
>

samphippen:
@JonRowe p sure the unicode ellipses literal won't work on 1.8.7, maybe it should just always be `"..."`

liam-m:
There were some CI failures with the ellipsis char in source

liam-m:
My intention with the unicode char was to be more semantically correct. I prefer how it looks, but that's just a preference. I don't mind much either way, happy to go with an agreement from other contributors. Leaning towards "..." now, given the output encoding may not be utf-8. 

Besides that, still unsure about:

* Whether the class methods should be all together or after each instance method, as they currently are
* The ```# private``` and ```# @api private``` comments

Also planning to add some comments about why the default instance and class method forwarding exists. 

Are any more tests needed?

samphippen:
@liam-m I've just gone over your implementation and I have a few concerns. You've changed `ObjectFormatter` from being a module to a class. Instead, why don't we do this, and keep it a module

```ruby
def self.max_formatted_output_length=(value)
  @max_formatted_output_length = value
end

def self.max_formatted_output_length
  #this defaults the output length to -1
  @max_formatted_output_length ||= -1
end 
```

then, you wouldn't need two implementations of each method and you wouldn't need to change the thing to being a class.

Also: both of those methods should be marked `#@api private`

Let's use `"..."` as our ellipses literal.

If the value is -1, it should print the whole output, and -1 should be the default.

Can you make those changes? Do you like this approach?

liam-m:
@samphippen, @myronmarston [suggested turning it into a class](https://github.com/rspec/rspec-support/pull/256#discussion_r47413508), to avoid adding mutable state at the module level, which is a code smell. 

Is there documentation on the ```# @api private``` comments? If not, could someone explain their use (just to mark as private?) and the difference between them and ```# private``` comments?

I'll make the change to ```"..."```

@myronmarston also suggested the default ```max_formatted_output_length``` [should be 200](https://github.com/rspec/rspec-support/pull/256#discussion_r47420000), i.e. by default the long output should be truncated. This makes sense to me.

samphippen:
Ok, cool, leaving it as a class is fine with me then :), Api private is a way of marking a method as not part of rspecs public api, and is used to let others know they shouldn't be calling those methods. 200 default length is fine.

liam-m:
Thanks for the quick response :). So should I add ```# @api private``` to all of the methods in the class?

samphippen:
@liam-m all methods on this class should be marked with `# @api private`

liam-m:
Some of them just have ```@private``` - is that okay? 

The failures on Travis are because the class is too long after adding the comments - now 101 lines and the limit is 100. Should I shorten it somehow, or up the limit in this case?

Beyond that, I'm still planning on adding comments about the method forwarding - I think it's now ready to do that as it's decided.

Are there any more tests you would like implementing?

lib/rspec/support/object_formatter.rb
```diff
@@ -2,10 +2,38 @@ module RSpec
   module Support
     # Provide additional output details beyond what `inspect` provides when
     # printing Time, DateTime, or BigDecimal
```
JonRowe: If you add `# @api private` here it will apply to all the methods in the class and you won't need the individual ones. This will neatly side step the rubocop issue.

liam-m: I also removed the individual ```# @private``` comments, please let me know if this wasn't the correct thing to do

liam-m:
@JonRowe @samphippen @myronmarston Would you like any more tests implementing for this? Is there anything else that needs doing? I assume we'll clean up the history before merging.

spec/rspec/support/object_formatter_spec.rb
```diff
@@ -158,6 +158,13 @@ def with_delegate_loaded
             matcher_without_a_description.inspect)
         end
       end
+
+      context 'with truncation' do
+        it 'produces an output of limited length' do
+          formatter = ObjectFormatter.new(10)
```
samphippen: I'd like to see a length of `-1` disable this behaviour, and print the full object. Could you implement and test that. Other than that, I think this is good to merge. Also: we need a test for an object with an inspect string that is shorter than this number. What happens here with a 9 character string?

liam-m: At the moment, nil disables it. Is -1 required? If so, should I remove the nil behaviour? Should it be -1 explicitly or any negative number?

samphippen: ah no, nil is fine. the -1 is less good than nil. Can you add a test for that case though? Also shorter inspect strings?

liam-m:
Shall I rebase this to one commit?

JonRowe:
Yes please

liam-m:
Ready to merge?

liam-m:
Wonderful! @JonRowe @samphippen @myronmarston thank you all for being so helpful during this process :)

PR #268, opened 2016-03-18
Fix JRuby bug with encoded string splitting

JonRowe:
JRuby has a different behaviour when splitting strings to all the other Rubies


myronmarston:
LGTM

Issue #267, opened 2016-03-05
Stack too deep when displaying failure message


blatyo:
I've constructed a simple reproduction of an issue I noticed. It involves the formatting of an error message, when comparing the equality of a recursive hash.

``` ruby
x = {}
x[:x] = x
puts x #=> {:x=>{...}}
expect(x).to eq({})
```

It gets stuck repeating these calls:
```
/Users/blatyo/.rvm/gems/ruby-2.2.3/gems/rspec-support-3.3.0/lib/rspec/support/object_formatter.rb:50:in `block in prepare_hash'
/Users/blatyo/.rvm/gems/ruby-2.2.3/gems/rspec-support-3.3.0/lib/rspec/support/object_formatter.rb:49:in `each'
/Users/blatyo/.rvm/gems/ruby-2.2.3/gems/rspec-support-3.3.0/lib/rspec/support/object_formatter.rb:49:in `inject'
/Users/blatyo/.rvm/gems/ruby-2.2.3/gems/rspec-support-3.3.0/lib/rspec/support/object_formatter.rb:49:in `prepare_hash'
/Users/blatyo/.rvm/gems/ruby-2.2.3/gems/rspec-support-3.3.0/lib/rspec/support/object_formatter.rb:28:in `prepare_for_inspection'
```

Full stack trace:
```
/Users/blatyo/.rvm/gems/ruby-2.2.3/gems/rspec-support-3.3.0/lib/rspec/support/object_formatter.rb:50:in `block in prepare_hash'
/Users/blatyo/.rvm/gems/ruby-2.2.3/gems/rspec-support-3.3.0/lib/rspec/support/object_formatter.rb:49:in `each'
/Users/blatyo/.rvm/gems/ruby-2.2.3/gems/rspec-support-3.3.0/lib/rspec/support/object_formatter.rb:49:in `inject'
/Users/blatyo/.rvm/gems/ruby-2.2.3/gems/rspec-support-3.3.0/lib/rspec/support/object_formatter.rb:49:in `prepare_hash'
/Users/blatyo/.rvm/gems/ruby-2.2.3/gems/rspec-support-3.3.0/lib/rspec/support/object_formatter.rb:28:in `prepare_for_inspection'
/Users/blatyo/.rvm/gems/ruby-2.2.3/gems/rspec-support-3.3.0/lib/rspec/support/object_formatter.rb:50:in `block in prepare_hash'
[...]

PR #269, opened 2016-03-19
Support objects that don't respond to #inspect in ObjectFormatter

yujinakayama:
I just found another bug in `ObjectFormatter` while working on #267.


JonRowe:
LGTM

PR #270, opened 2016-03-19
Support recursive collections in ObjectFormatter

yujinakayama:
This fixes #267.

PR #272, opened 2016-03-22
Handle non-immediate recursive structures in ObjectFormatter

yujinakayama:
Pointed out in: https://github.com/rspec/rspec-support/pull/270#issuecomment-199298121

### DIFF
--- a/lib/rspec/support/object_formatter.rb
+++ b/lib/rspec/support/object_formatter.rb
@@ -5,7 +5,7 @@ module RSpec
     # Provide additional output details beyond what `inspect` provides when
     # printing Time, DateTime, or BigDecimal
     # @api private
-    class ObjectFormatter
+    class ObjectFormatter # rubocop:disable Style/ClassLength
       ELLIPSIS = "..."
 
       attr_accessor :max_formatted_output_length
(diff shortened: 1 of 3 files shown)


lib/rspec/support/object_formatter.rb
```diff
@@ -65,30 +66,44 @@ def prepare_for_inspection(object)
...
+      end
 
-          output_hash[key] = value
-          output_hash
+      def prepare_element(element)
+        if recursive_structure?(element)
+          case element
+          when Array then InspectableItem.new('[...]')
+          when Hash then InspectableItem.new('{...}')
+          else raise
```
myronmarston: If we're going to raise an error, it should at least have a message, although I'm not sure that's what we want.

JonRowe: I'd do `InspectableItem.new('<...>')` to indicate a recursive object?

yujinakayama: Actually the `raise` is a sort of assertion and will never be run since `#recursive_structure?` returns true only when a hash or array is given.

JonRowe:
:heart:

PR #273, opened 2016-04-15
RSpec::Support::FuzzyMatcher.failure_message implemented…

faucct:
… for providing informative failure messages

### DIFF
--- a/lib/rspec/support/fuzzy_matcher.rb
+++ b/lib/rspec/support/fuzzy_matcher.rb
@@ -3,46 +3,99 @@ module RSpec
     # Provides a means to fuzzy-match between two arbitrary objects.
     # Understands array/hash nesting. Uses `===` or `==` to
     # perform the matching.
-    module FuzzyMatcher
-      # @api private
-      def self.values_match?(expected, actual)
+    class FuzzyMatcher
+      def initialize(expected, actual)
+        @failure_path, @raw_failure_message = failure_path_and_message(expected, actual) || []
+        @failure_message =
+          if @raw_failure_message
+            inspected_path = "#{actual.inspect}#{@failure_path}"
+            if @raw_failure_message.empty?
+              "#{inspected_path} failed to match"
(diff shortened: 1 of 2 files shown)


myronmarston:
What are you trying to achieve here?  `FuzzyMatcher` is not an official rspec-expectations matcher, and isn't meant to be used directly by end users.  Can you show an example of what you wanted to do (but could not) that this PR would fix?

faucct:
Hello, I want to make the rspec-core's `include_attributes` failure_message more descriptive and it utilizes the `FuzzyMatcher`.

myronmarston:
Sounds like the same issue of rspec/rspec-expectations#759.  Does that sound right?

Regardless, any improvements to the `have_attributes` matcher are going to have to be made in rspec-expectations, since that's where it lives.   `FuzzyMatcher` can expose some additional APIs to support corresponding changes over there as needed of course.  Can you open a PR showing what you have in mind for the changes to `have_attributes`?

lib/rspec/support/fuzzy_matcher.rb
```diff
@@ -3,46 +3,100 @@ module Support
...
 
-        return true if expected == actual
+        return if expected == actual
 
         begin
-          expected === actual
+          if expected === actual
+            return
+          else
+            if Matchers::BuiltIn::BaseMatcher === expected
```
myronmarston: RSpec::Support must be usable w/o rspec-expectations loaded (which is the gem that defines `Matchers::BuiltIn::BaseMatcher`) so doing this kind of check here is problematic.

faucct:
I am back with cleaned up code. The `have_attributes` PR example will take some time, I will do it a little bit later.

JonRowe:
:wave: I'm closing this PR as stale for now, it also makes too many breaking changes to `FuzzyMatcher` any improvements really need to be focused on the matcher in question. You certainly can't change this to a class.

PR #275, opened 2016-04-22
fix sometimes no HOME path set

hendricius:
I received an error when deploying the latest master to AWS Beanstalk. In the deployment process no `HOME` ENV variable was set.

```
There was a ArgumentError while loading rspec-support.gemspec:
couldn't find HOME environment -- expanding `~' from
/opt/rubies/ruby-2.3.0/lib/ruby/gems/2.3.0/bundler/gems/rspec-support-d8666d5fc14a/rspec-support.gemspec:22:in
`expand_path'.
```

This pull request fixes that. IF you have feedback, please let me know.

### DIFF
--- a/rspec-support.gemspec
+++ b/rspec-support.gemspec
@@ -19,7 +19,12 @@ Gem::Specification.new do |spec|
   spec.rdoc_options  = ["--charset=UTF-8"]
   spec.require_paths = ["lib"]
 
-  private_key = File.expand_path('~/.gem/rspec-gem-private_key.pem')
+
+  private_key = ""
+  begin
+    private_key = File.expand_path('~/.gem/rspec-gem-private_key.pem')
+  rescue ArgumentError => e
+  end
   if File.exist?(private_key)
     spec.signing_key = private_key
     spec.cert_chain = [File.expand_path('~/.gem/rspec-gem-public_cert.pem')]


rspec-support.gemspec
```diff
@@ -19,7 +19,12 @@ Gem::Specification.new do |spec|
   spec.rdoc_options  = ["--charset=UTF-8"]
   spec.require_paths = ["lib"]
 
-  private_key = File.expand_path('~/.gem/rspec-gem-private_key.pem')
+
+  private_key = ""
+  begin
+    private_key = File.expand_path('~/.gem/rspec-gem-private_key.pem')
+  rescue ArgumentError => e
+  end
```
JonRowe: It would be better to wrap all of the signing logic in this block, rather than run a useless (and potentially wrong) if check,  also maybe it's worth checking for the presence of HOME instead rather than rescuing a misleading exception, WDYT

JonRowe:
This will need to be done for all our gems if we go ahead with it, however have you considered that you shouldn't be deploying test gems? Normally people deploy `bundler --deployment` which skips gems like rspec.

hendricius:
Not sure if we still need this. We are no longer using AWS ourselves. I will close this one.

PR #281, opened 2016-05-24
Ignore bundler warnings on windows

JonRowe:
Currently windows builds get broken by a bundler warning

### DIFF
--- a/lib/rspec/support/spec/library_wide_checks.rb
+++ b/lib/rspec/support/spec/library_wide_checks.rb
@@ -68,9 +68,7 @@ RSpec.shared_examples_for "library wide checks" do |lib, options|
       run_ruby_with_current_load_path(command, *options)
     end
 
-    # Ignore bundler warning.
-    stderr = stderr.split("\n").reject { |l| l =~ %r{bundler/source/rubygems} }.join("\n")
-    [stdout, stderr, status.exitstatus]
+    [stdout, strip_known_warnings(stderr), status.exitstatus]
   end
 
   define_method :load_all_lib_files do
--- a/spec/rspec/support/spec/shell_out_spec.rb
+++ b/spec/rspec/support/spec/shell_out_spec.rb
@@ -33,7 +33,7 @@ RSpec.describe RSpec::Support::ShellOut, :slow do
   it 'passes along the provided ruby flags' do
     out, err, status = run_ruby_with_current_load_path('puts "version"', '-v')
     expect(out).to include('version', RUBY_DESCRIPTION)
-    expect(err).to eq('')
+    expect(strip_known_warnings err).to eq('')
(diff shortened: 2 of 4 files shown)


lib/rspec/support/spec/library_wide_checks.rb
```diff
@@ -68,8 +68,13 @@ def load_all_files(files, preamble, postamble=nil)
...
     end
 
-    # Ignore bundler warning.
-    stderr = stderr.split("\n").reject { |l| l =~ %r{bundler/source/rubygems} }.join("\n")
+    stderr = stderr.split("\n").reject do |l|
+      # Ignore bundler warning.
+      l =~ %r{bundler/(source/)?rubygems} ||
+      # Ignore bundler + rubygems warning.
+      l =~ %r{site_ruby/\d\.\d\.\d/rubygems} ||
+      l =~ %r{lib/bundler/rubygems}
```
myronmarston: It looks like the `l =~ %r{bundler/(source/)?rubygems}` line will match anything this regex was match -- so do we really need it?

spec/rspec/support/spec/shell_out_spec.rb
```diff
@@ -33,16 +33,26 @@
...
+
+  def ignore_bundler_warnings(input)
+    input.split("\n").reject do |l|
+      # Ignore bundler warning.
+      l =~ %r{bundler/(source/)?rubygems} ||
+      # Ignore bundler + rubygems warning.
+      l =~ %r{site_ruby/\d\.\d\.\d/rubygems} ||
+      l =~ %r{lib/bundler/rubygems}
+    end.join("\n")
+  end
```
myronmarston: There's a lot of duplication between this and the other file.  Can this method be pulled into a helper module?

lib/rspec/support/spec/library_wide_checks.rb
JonRowe: I'm not sure why the first regex doesn't work, but it doesn't :/

myronmarston: What's an example of a string that matches this but not the first?

JonRowe: `lib/bundler/rubygems` I only added it because the regex wasn't catching that on windows

myronmarston: Rubular reports that that matches:

http://rubular.com/r/T5HRERvB54

Do regexes work differently when MRI is compiled on windows?

Anyhow, can you add a comment explaining why this extra regex is needed even though it looks like it shouldn't be?

JonRowe:
Updayed.

myronmarston:
LGTM, merge when green.

PR #294, opened 2016-08-26
Bugfix - Remove partial ANSI colour codes when truncating object.inspect

ansonK:
When truncating an object using `ObjectFormatter.format` that exceeds the max_formatted_output_length remove any partial ANSI color codes.

This fixes an edge case where partial ANSI color codes eg `\e[3` is printed to the terminal which corrupts the output.

eg calling object.inspect = "#<\e[33mClass\e[0m \e[36mname: \e[0m\"foobars\" \e[36mcount: \e[0m42>"
If this was truncated at the 19th char then the result would be: "#<\e[33mClass\e[0m \e[3"
(which results in an invalid escape code).
This commit will remove the partial ANSI code so the result is now: "#<\e[33mClass\e[0m "

To demonstrate the problem here is a screenshot of an example spec running rspec-support 3.4 with good output:
screen shot 2016-08-26 at 15 44 13

And here is the same spec running rspec-support 3.5 with a bad ANSI code corrupting output:
(The output in the `expected [] to include...` line contains the bad ANSI code which seems to reset the terminal to the top so the output overwrites)
screen shot 2016-08-26 at 15 42 47

### DIFF
--- a/lib/rspec/support/object_formatter.rb
+++ b/lib/rspec/support/object_formatter.rb
@@ -37,8 +37,8 @@ module RSpec
           if formatted_object.length < max_formatted_output_length
             return formatted_object
           else
-            beginning = formatted_object[0 .. max_formatted_output_length / 2]
-            ending = formatted_object[-max_formatted_output_length / 2 ..-1]
+            beginning = truncate_string formatted_object, 0, max_formatted_output_length / 2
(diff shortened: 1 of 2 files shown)


JonRowe:
Thanks :heart:

PR #297, opened 2016-10-21
Allow introspection into ReentrantMutex

drcapulet:
Required for rspec/rspec-core#2340.

### DIFF
--- a/lib/rspec/support/reentrant_mutex.rb
+++ b/lib/rspec/support/reentrant_mutex.rb
@@ -25,10 +25,18 @@ module RSpec
         exit
       end
 
+      def locked?
+        @owner != nil
+      end
+
+      def owned?
+        @owner == Thread.current
+      end
+
     private
 
       def enter
-        @mutex.lock if @owner != Thread.current
+        @mutex.lock unless owned?
         @owner = Thread.current
         @count += 1
       end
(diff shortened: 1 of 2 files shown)


lib/rspec/support/reentrant_mutex.rb
```diff
@@ -25,10 +25,18 @@ def synchronize
         exit
       end
 
+      def locked?
+        @owner != nil
+      end
+
+      def owned?
+        @owner == Thread.current
+      end
```
drcapulet: Updated, let me know if you want anything more complex.

JonRowe: Great, as per @joshcheek's comments on rspec/rspec-core#2340 we still need these methods on our own version of mutex for 1.8.7 too :)

drcapulet: `ReentrantMutex` is a wrapper on top of either the native Ruby `Mutex` class of the RSpec version ([initialization](https://github.com/rspec/rspec-support/blob/master/lib/rspec/support/reentrant_mutex.rb#L18) and [definition](https://github.com/rspec/rspec-support/blob/master/lib/rspec/support/reentrant_mutex.rb#L44-L51)) so this should be the only place changes are required.

JonRowe:
Thanks :)

PR #298, opened 2016-11-03
Add configuration for max object formatter length.

samphippen:
See: https://github.com/rspec/rspec-expectations/pull/951

### DIFF
--- a/lib/rspec/support/object_formatter.rb
+++ b/lib/rspec/support/object_formatter.rb
@@ -16,6 +16,10 @@ module RSpec
         @default_instance ||= new
       end
 
+      def self.default_max_formatted_output_length=(length)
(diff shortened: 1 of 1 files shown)


lib/rspec/support/object_formatter.rb
```diff
@@ -16,6 +16,10 @@ def self.default_instance
         @default_instance ||= new
       end
 
+      def self.default_max_formatted_output_length=(length)
+        default_instance.max_formatted_output_length = length
```
JonRowe: I think it's fine to call this directly from expectations, both are private anyway.

samphippen: Totally. I did this to have a shorter call chain over on the Expectations side. Do you think this is fine and we can just merge?

PR #310, opened 2017-03-22
Define `supports_qualified_constant_toplevel_lookup?`

yui-knk:
In Ruby 2.5, top-level constant look-up through a qualified constant
is not permitted, so we have to check qualified constants top level
look up is supported.

see: https://github.com/ruby/ruby/commit/44a2576f798b07139adde2d279e48fdbe71a0148

### DIFF
--- a/lib/rspec/support/ruby_features.rb
+++ b/lib/rspec/support/ruby_features.rb
@@ -90,6 +90,23 @@ module RSpec
         end
       end
 
+      begin
+        verbose_was = $VERBOSE
+        $VERBOSE = nil
+
+        _ = String::Hash
+
+        def supports_qualified_constant_toplevel_lookup?
+          true
+        end
+      rescue NameError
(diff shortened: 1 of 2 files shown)


lib/rspec/support/ruby_features.rb
```diff
@@ -90,6 +90,23 @@ def ripper_supported?
         end
       end
 
+      begin
+        verbose_was = $VERBOSE
+        $VERBOSE = nil
+
+        String::Hash
```
samphippen: This is failing on windows because of the void context of this lookup. I think assigning this to a local that gets thrown away would solve it 👍

JonRowe:
:wave: apologies in the delay in getting back to you, this seems to work now which is great, but I'm a bit confused as to where it would be used?

pirj:
Closing as stale

PR #313, opened 2017-04-16
Reflect Ripper support in the latest JRuby in #ripper_supported

yujinakayama:
... so that rspec-core on JRuby 9.1.0.0 or later can benefit from the multiline failure snippet extraction.

I've checked if the spec passes on all JRuby 9.x.x.x.

https://travis-ci.org/rspec/rspec-support/builds/222465272

screen shot 2017-04-16 at 9 42 51

PR #314, opened 2017-04-18
Mark Ripper on the latest JRuby as broken

yujinakayama:
It turned out that Ripper on the latest JRuby still had [another bug](https://github.com/jruby/jruby/issues/4562).

This fixes CI failure reported in https://github.com/rspec/rspec-support/pull/313#issuecomment-294385384.


JonRowe:
Thanks for following up @yujinakayama !

PR #315, opened 2017-05-03
Import source classes from rspec-core

yujinakayama:
For https://github.com/rspec/rspec-expectations/pull/987.

This corresponds to https://github.com/rspec/rspec-core/pull/2420.

### DIFF
--- a/lib/rspec/support/source/location.rb
+++ b/lib/rspec/support/source/location.rb
@@ -4,8 +4,16 @@ module RSpec
       # @private
       # Represents a source location of node or token.
       Location = Struct.new(:line, :column) do
+        include Comparable
+
         def self.location?(array)
           array.is_a?(Array) && array.size == 2 && array.all? { |e| e.is_a?(Integer) }
+        end
+
+        def <=>(other)
+          line_comparison = (line <=> other.line)
+          return line_comparison unless line_comparison == 0
+          column <=> other.column
         end
       end
     end
--- a/lib/rspec/support/source/location.rb
+++ b/lib/rspec/support/source/location.rb
@@ -1,0 +1,13 @@
+module RSpec
+  module Support
+    class Source
+      # @private
+      # Represents a source location of node or token.
+      Location = Struct.new(:line, :column) do
+        def self.location?(array)
+          array.is_a?(Array) && array.size == 2 && array.all? { |e| e.is_a?(Integer) }
+        end
+      end
+    end
+  end
+end
--- a/lib/rspec/support/source/node.rb
+++ b/lib/rspec/support/source/node.rb
@@ -30,7 +30,7 @@ module RSpec
             elsif Location.location?(raw_arg)
               Location.new(*raw_arg)
             elsif raw_arg.is_a?(Array)
-              GroupNode.new(raw_arg, self)
+              ExpressionSequenceNode.new(raw_arg, self)
             else
               raw_arg
             end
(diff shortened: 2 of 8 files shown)


lib/rspec/support/source/node.rb
```diff
@@ -0,0 +1,244 @@
...
+          :word_new,
+          :words_add,
+          :words_new,
+          :xstring_add,
+          :xstring_literal,
+          :xstring_new,
+          :yield,
+          :yield0,
+          :zsuper
+        ]
```
myronmarston: I worry that this array will get out of sync with the ripper constant.  It looks like it's only needed to dynamically define all the `*_type?` methods.  Instead, WDYT about defining  a single `of_type?(event)` method or just do manual `==` comparisons?  Then you wouldn't need this constant, this class would be smaller/simpler, and we wouldn't have to worry about the constant going out of sync with Ripper's.

On a side note: the rspec-core `Node` class did not need this, so we do we need it now?

lib/rspec/support/source/node.rb
```diff
@@ -0,0 +1,244 @@
...
+
+      private
+
+        def raw_args
+          sexp[1..-1] || []
+        end
+      end
+
+      # @private
+      class ExpressionSequenceNode < Node
```
myronmarston: I noticed rspec-core had `GroupNode` vs here you have `ExpressionSequenceNode`.  Can you explain why it needed to change?

BTW, a comment in the source explaining the need for this subclass would be helpful to our future selves, I think.

lib/rspec/support/source/token.rb
```diff
@@ -0,0 +1,149 @@
...
+          :symbeg,
+          :symbols_beg,
+          :tlambda,
+          :tlambeg,
+          :tstring_beg,
+          :tstring_content,
+          :tstring_end,
+          :words_beg,
+          :words_sep
+        ].freeze
```
myronmarston: Here too I'd like to see us get away from a constant we have to keep in sync with ripper...and I'm also curious as to why we need it here but did not in rspec-core.

lib/rspec/support/source/location.rb
```diff
@@ -0,0 +1,21 @@
...
+
+        def self.location?(array)
+          array.is_a?(Array) && array.size == 2 && array.all? { |e| e.is_a?(Integer) }
+        end
+
+        def <=>(other)
+          line_comparison = (line <=> other.line)
+          return line_comparison unless line_comparison == 0
+          column <=> other.column
+        end
```
yujinakayama: It's needed here https://github.com/rspec/rspec-expectations/pull/987/commits/c074ca900790336204ae3e18e88c03ebf337ce24#diff-9b277082395427eb0159d15a3204096bR176.

lib/rspec/support/source/node.rb
yujinakayama: This was for convenience:

* `node.select { |node| node.type == :fcall }` can be described as `node.select(&:fcall_type?)`
* We can notice if there's a typo in type name (`node.fcalll_type? # NoMethodError`)

> WDYT about defining a single `of_type?(event)` method or just do manual `==` comparisons?

Will do.

lib/rspec/support/source/node.rb
yujinakayama: It's for clarity, since _group_ was too generic word.

yujinakayama: > BTW, a comment in the source explaining the need for this subclass would be helpful to our future selves, I think.

Done

PR #316, opened 2017-05-25
Fixing strings if they're frozen literals.

pat:
I realise += isn't great from a performance perspective, but this only crops up with warnings/deprecations.


samphippen:
👍

PR #317, opened 2017-05-27
Explicitly get a module name using `name`.

myronmarston:
Instead of letting `Array#join` try to use `to_str` on the module.

This works around buggy modules/classes that raise odd errors from
`to_str`. This fixes rspec/rspec-mocks#1161.


JonRowe:
:+1:

PR #321, opened 2017-06-28
object_formatter_spec.rb: fix tests on ia64

trofi:
The error is caused by formatter:

```ruby
    class UninspectableObjectInspector < BaseInspector
        OBJECT_ID_FORMAT = '%#016x'
    ...
```

'%#016x' produces 16 bytes most of the time (including leading "0x").
Unless original address is more than 7 bytes.

ia64 architecture happens to use highest bytes extensively:

```
$ cat /proc/self/maps
00000000-00004000 r--p 00000000 00:00 0
2000000000000000-2000000000044000 r-xp 00000000 68:04 5205554            /lib/ld-2.24.so
2000000000044000-200000000004c000 rw-p 00000000 00:00 0
```

Note 16-byte addresses. That makes tests fail as:

```
Failures:

  1) RSpec::Support::ObjectFormatter.format with an object that does not respond to #inspect such as BasicObject produces an output emulating MRI's #inspect-like output generated by C impleme
ntation
     Failure/Error: expect(output).to match(/\A#<BasicObject:0x[0-9a-f]{14}>\z/)

       expected "#<BasicObject:0x6000000001466b10>" to match /\A#<BasicObject:0x[0-9a-f]{14}>\z/
       Diff:
       @@ -1,2 +1,2 @@
       -/\A#<BasicObject:0x[0-9a-f]{14}>\z/
       +"#<BasicObject:0x6000000001466b10>"

     # ./spec/rspec/support/object_formatter_spec.rb:200:in `block (3 levels) in <module:Support>'

Finished in 1.53 seconds (files took 0.75756 seconds to load)
475 examples, 1 failure, 2 pending

```

Signed-off-by: Sergei Trofimovich <slyfox@gentoo.org>


myronmarston:
Thanks for the fix for a really subtle issue!

PR #334, opened 2017-11-24
Fix diffs for hashes inside arrays, by ordering keys

jcrisp:
Hi! 

I noticed diffs were sometimes incorrect when mocking calls and checking hash arguments. See example specs & output here:
https://github.com/jcrisp/rspec-mocks/commit/28e88c4a6cd8f3887c42b0beefa4a3312a120230 

On digging further, it looked best to fix this in rspec-support by ordering the keys in hashes when they are nested in arrays. Happy to improve the PR or redo elsewhere in rspec if you'd prefer.

Cheers
James

### DIFF
--- a/lib/rspec/support/object_formatter.rb
+++ b/lib/rspec/support/object_formatter.rb
@@ -73,11 +73,19 @@ module RSpec
 
       def prepare_hash(input_hash)
         with_entering_structure(input_hash) do
-          input_hash.inject({}) do |output_hash, key_and_value|
+          sort_hash_keys(input_hash).inject({}) do |output_hash, key_and_value|
             key, value = key_and_value.map { |element| prepare_element(element) }
             output_hash[key] = value
             output_hash
           end
+        end
+      end
+
+      def sort_hash_keys(input_hash)
+        if input_hash.keys.all? { |k| k.is_a?(String) || k.is_a?(Symbol) }
+          Hash[input_hash.sort_by { |k, _v| k.to_s }]
+        else
+          input_hash
         end
       end
 
--- a/spec/rspec/support/differ_spec.rb
+++ b/spec/rspec/support/differ_spec.rb
@@ -228,6 +228,27 @@ EOD
           expect(diff).to be_diffed_as(expected_diff)
         end
 
+        unless RUBY_VERSION == '1.8.7' # We can't count on the ordering of the hash on 1.8.7...
+          it "outputs unified diff message for hashes inside arrays with differing key orders" do
+            expected = [{ :foo => 'bar', :baz => 'quux', :metasyntactic => 'variable', :delta => 'charlie', :width =>'quite wide' }]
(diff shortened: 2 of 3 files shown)


jcrisp:
Here's the PR to fix the 3 failing tests in rspec-expectations (ordering of keys in error message)
https://github.com/rspec/rspec-expectations/pull/1027

jcrisp:
@JonRowe Jon, I needed to also make a very small change in rspec-mocks specs to order the keys (see PR https://github.com/rspec/rspec-mocks/pull/1189). The fix works fine everywhere except Ruby 1.8.7, where hash key order is not preserved. For Ruby 1.8.7 people have the same diff behaviour as before, but all other (more modern) versions they get the improved diffs.

Just to summarise, to merge this change, it needs this PR and two other tiny supporting PRs which change key order in spec expectations:
https://github.com/rspec/rspec-mocks/pull/1189
https://github.com/rspec/rspec-expectations/pull/1027

The first commit in this PR is the actual change.
The second commit is just the pinning against my others PRs so the build can pass.

Thanks
James

lib/rspec/support/object_formatter.rb
```diff
@@ -73,14 +73,22 @@ def prepare_array(array)
...
             output_hash
           end
         end
       end
 
+      def sort_hash_keys(input_hash)
+        if input_hash.keys.all? { |k| k.is_a?(String) || k.is_a?(Symbol) }
+          Hash[input_hash.sort_by { |k, _v| k.to_s }]
+        else
+          input_hash
```
xaviershay: I don't think this branch is tested?

script/clone_all_rspec_repos
```diff
@@ -10,8 +10,8 @@ if is_mri; then
 
   clone_repo "rspec"
   clone_repo "rspec-core"
-  clone_repo "rspec-expectations"
-  clone_repo "rspec-mocks"
+  clone_repo "rspec-expectations" "jcrisp"
+  clone_repo "rspec-mocks" "jcrisp"
```
xaviershay: Need to remember to revert this before merging

lib/rspec/support/object_formatter.rb
xaviershay: Oh actually it's the default so probably covered by existing

jcrisp: Hi @xaviershay! Yes the raw sorting of keys is covered in object_formatter_spec.rb:38 and differ_spec.rb:248 makes sure that the diff is right. Without the sorting, there'll be incorrect additional keys in the diff which will fail differ_spec.rb:248.

JonRowe:
Ok this looks good, you can remove the repo in and we'll merge it!

jcrisp:
Cheers @JonRowe and @xaviershay, I've removed the other repo pinning commit, should be good to go.

PR #342, opened 2018-01-26
Add `fork_supported?`.

myronmarston:
This is needed for the new fork-based bisect runner
in rspec/rspec-core#2511.

### DIFF
--- a/lib/rspec/support/ruby_features.rb
+++ b/lib/rspec/support/ruby_features.rb
@@ -56,6 +56,22 @@ module RSpec
     module RubyFeatures
       module_function
 
+      if Ruby.jruby?
+        # On JRuby 1.7 `--1.8` mode, `Process.respond_to?(:fork)` returns true,
+        # but when you try to fork, it raises an error:
+        #   NotImplementedError: fork is not available on this platform
+        #
+        # When we drop support for JRuby 1.7 and/or Ruby 1.8, we can drop
+        # this special case.
+        def fork_supported?
+          false
+        end
+      else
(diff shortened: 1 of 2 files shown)


myronmarston:
Can I get a review of this, @rspec/rspec ?  I tried to target the Gemfile in rspec/rspec-core#2511 at this branch so I can use this change there but it's not working for some reason.  Merging this is the simplest path forward.

lib/rspec/support/ruby_features.rb
```diff
@@ -56,6 +56,22 @@ def mri?
...
+        # but when you try to fork, it raises an error:
+        #   NotImplementedError: fork is not available on this platform
+        #
+        # When we drop support for JRuby 1.7 and/or Ruby 1.8, we can drop
+        # this special case.
+        def fork_supported?
+          false
+        end
+      else
+        def fork_supported?
```
JonRowe: Could we define this conditionally or memoized to avoid calling `Process.respond_to?(:fork)` repeatedly?

myronmarston: I'd want a benchmark showing memoization is faster as `Process.respond_to?(:fork)` isn't something I'd expect to be slow.  Re: defining it conditionally...we could, but that's not an approach we're consistently following in this file.  If we want to go that route, I think we should do it consistently throughout this file.

This method is only going to be called once per spec run (in `RSpec::Core::Configuration#initialize`) so it's definitely not a perf hot spot, anyway.

PR #344, opened 2018-02-15
Use https:// instead of git:// URLs in READMEs

alyssais:
See https://github.com/rspec/rspec-core/pull/2518.


myronmarston:
Thanks!

PR #345, opened 2018-03-01
Workaround JRuby incompatibilities

grddev:
JRuby apparently only supports #arity (and not #parameters) for Java proxy methods (see https://github.com/jruby/jruby/issues/2817), but verifying doubles in RSpec use #parameters to infer required method parameters, and thus doesn't work for Java classes.

This extends the previously existing JRuby workaround that returns an empty list from #parameters so that it fallbacks to the legacy method of using #arity.

In addition, it turns out that the #arity method is broken for Java proxy methods in JRuby 1.7.x. While JRuby 1.7 has been sunset, the test suite still uses it, and thus this introduces another workaround to account for that.

This looks up the corresponding JavaMethod overloads, and if there is only one, retrieves the arity from that one instead. In the presence of multiple overloads, -1 is returned.

While this obviously doesn't cover all cases, it does cover the cases the previous workaround covered, and also allows using verifying doubles for many Java classes in JRuby.

### DIFF
--- a/lib/rspec/support/method_signature_verifier.rb
+++ b/lib/rspec/support/method_signature_verifier.rb
@@ -154,25 +154,56 @@ module RSpec
       INFINITY = 1 / 0.0
     end
 
-    # JRuby has only partial support for UnboundMethod#parameters, so we fall back on using #arity
-    # https://github.com/jruby/jruby/issues/2816 and https://github.com/jruby/jruby/issues/2817
-    if RSpec::Support::Ruby.jruby? &&
-       RubyFeatures.optional_and_splat_args_supported? &&
-       Java::JavaLang::String.instance_method(:char_at).parameters == []
+    if RSpec::Support::Ruby.jruby?
+      # JRuby has only partial support for UnboundMethod#parameters, so we fall back on using #arity
+      # https://github.com/jruby/jruby/issues/2816 and https://github.com/jruby/jruby/issues/2817
+      if RubyFeatures.optional_and_splat_args_supported? &&
+         Java::JavaLang::String.instance_method(:char_at).parameters == []
 
-      class MethodSignature < remove_const(:MethodSignature)
-      private
+        class MethodSignature < remove_const(:MethodSignature)
+        private
 
-        def classify_parameters
-          super
-          if (arity = @method.arity) != 0 && @method.parameters.empty?
-            if arity < 0
-              @min_non_kw_args = ~arity
-              @max_non_kw_args = INFINITY
-            else
-              @min_non_kw_args = arity
-              @max_non_kw_args = arity
+          def classify_parameters
+            super
+            if (arity = @method.arity) != 0 && @method.parameters.empty?
+              if arity < 0
+                @min_non_kw_args = ~arity
+                @max_non_kw_args = INFINITY
+              else
+                @min_non_kw_args = arity
+                @max_non_kw_args = arity
+              end
             end
+          end
+        end
+      end
+
+      # JRuby used to always report -1 arity for Java proxy methods
+      if Java::JavaLang::String.instance_method(:char_at).arity == -1
+        class MethodSignature < remove_const(:MethodSignature)
+        private
+
+          def classify_parameters
+            super
+            if @method.arity == -1
+              arity = java_single_overload_arity
+              if arity < 0
+                @min_non_kw_args = ~arity
+                @max_non_kw_args = INFINITY
+              else
+                @min_non_kw_args = arity
+                @max_non_kw_args = arity
+              end
+            end
+          end
+
+          def java_single_overload_arity
+            return @method.arity unless @method.owner.respond_to?(:java_class)
+            java_instance_methods = @method.owner.java_class.java_instance_methods
+            compatible_overloads = java_instance_methods.select do |java_method|
+              @method == @method.owner.instance_method(java_method.name)
+            end
+            (compatible_overloads.size == 1) ? compatible_overloads.first.arity : @method.arity
           end
         end
       end
--- a/lib/rspec/support/method_signature_verifier.rb
+++ b/lib/rspec/support/method_signature_verifier.rb
@@ -154,19 +154,26 @@ module RSpec
       INFINITY = 1 / 0.0
     end
 
-    # Some versions of JRuby have a nasty bug we have to work around :(.
-    # https://github.com/jruby/jruby/issues/2816
+    # JRuby has only partial support for UnboundMethod#parameters, so we fall back on using #arity
+    # https://github.com/jruby/jruby/issues/2816 and https://github.com/jruby/jruby/issues/2817
     if RSpec::Support::Ruby.jruby? &&
        RubyFeatures.optional_and_splat_args_supported? &&
-       Class.new { attr_writer :foo }.instance_method(:foo=).parameters == []
+       Java::JavaLang::String.instance_method(:char_at).parameters == []
 
       class MethodSignature < remove_const(:MethodSignature)
       private
 
         def classify_parameters
           super
-          return unless @method.parameters == [] && @method.arity == 1
-          @max_non_kw_args = @min_non_kw_args = 1
+          if (arity = @method.arity) != 0 && @method.parameters.empty?
+            if arity < 0
+              @min_non_kw_args = ~arity
+              @max_non_kw_args = INFINITY
+            else
+              @min_non_kw_args = arity
+              @max_non_kw_args = arity
+            end
+          end
         end
       end
     end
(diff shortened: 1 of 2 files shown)


lib/rspec/support/method_signature_verifier.rb
```diff
@@ -154,19 +154,57 @@ def classify_parameters
...
-          return unless @method.parameters == [] && @method.arity == 1
-          @max_non_kw_args = @min_non_kw_args = 1
+          if @method.arity == -1 && (arity = java_method_arity)
+            if arity < 0
+              @min_non_kw_args = ~arity
+              @max_non_kw_args = INFINITY
+            else
+              @min_non_kw_args = arity
+              @max_non_kw_args = arity
+            end
```
myronmarston: These 6 lines (from `if arity < 0` to the `end`) show up 3 times in this file.  Can you extract a helper method (something like `classify_arity(arity)`) for this code so it's duplicated?

lib/rspec/support/method_signature_verifier.rb
```diff
@@ -154,19 +154,57 @@ def classify_parameters
...
+          end
+        end
+      end
+    end
 
+    if RSpec::Support::Ruby.jruby? &&
```
myronmarston: Instead of starting a 2nd conditional that starts with `if RSpec::Support::Ruby.jruby?`, would you mind doing one big `if RSpec::Support::Ruby.jruby?` conditional (with 2 nested conditionals as needed).

lib/rspec/support/method_signature_verifier.rb
```diff
@@ -154,19 +154,57 @@ def classify_parameters
...
     end
 
-    # Some versions of JRuby have a nasty bug we have to work around :(.
-    # https://github.com/jruby/jruby/issues/2816
+    # JRuby has only partial support for UnboundMethod#parameters, so we fall back on using #arity
+    # https://github.com/jruby/jruby/issues/2816 and https://github.com/jruby/jruby/issues/2817
     if RSpec::Support::Ruby.jruby? &&
        RubyFeatures.optional_and_splat_args_supported? &&
-       Class.new { attr_writer :foo }.instance_method(:foo=).parameters == []
+       Java::JavaLang::String.instance_method(:char_at).parameters == []
```
myronmarston: I'm trying to understand why you changed this conditional, and how this changes things.  Does `Class.new { attr_writer :foo }.instance_method(:foo=).parameters == []` return the same value on all versions of JRuby that `Java::JavaLang::String.instance_method(:char_at).parameters == []` does?

lib/rspec/support/method_signature_verifier.rb
myronmarston: Actually, is it possible to combine the two Jruby cases into one class, with some extra conditional checked when `classify_parameters` is called? It's confusing (and will be harder to maintain) to have two different alternate JRuby implementations.

lib/rspec/support/method_signature_verifier.rb
```diff
@@ -154,19 +154,57 @@ def classify_parameters
...
+              @max_non_kw_args = INFINITY
+            else
+              @min_non_kw_args = arity
+              @max_non_kw_args = arity
+            end
+          end
+        end
+
+        def java_method_arity
+          if @method.owner.respond_to?(:java_class)
```
myronmarston: I'd prefer to use an early-exit guard, like so:

``` ruby
return unless @method.owner.respond_to?(:java_class)
```

lib/rspec/support/method_signature_verifier.rb
```diff
@@ -154,19 +154,57 @@ def classify_parameters
...
+              @min_non_kw_args = ~arity
+              @max_non_kw_args = INFINITY
+            else
+              @min_non_kw_args = arity
+              @max_non_kw_args = arity
+            end
+          end
+        end
+
+        def java_method_arity
```
myronmarston: I don't totally get what you're doing in this method and why.  Mind leaving a comment explaining it?

grddev:
Thanks for the review @myronmarston. I think I could have done a better job explaining the problem in the PR description, so let me make another attempt. I wanted to use an instance double for a Java proxy in JRuby. For those not familiar with JRuby, you can freely interface with any standard Java class through an automatically generated proxy. Briefly, you can write `Java::JavaLang::String` in JRuby to access a Ruby-like proxy for the Java class `java.lang.String`, and you can use `#instance_method(:char_at)` to look up the Java method `java.lang.String#charAt(int)`.

However, the implementation of method signature information for such proxies are not stellar. Until 9.0.3.0 (and https://github.com/jruby/jruby/commit/013e288ade4d2c2e9b782d8e30f0f4b6df3157a0), `Method#arity` would always return -1 for these methods, and `Method#parameters` still returns an empty list independent of the underlying method(s), for all JRuby versions released so far. The following table summarizes the quality of the method signature helpers for different JRuby versions and command-line flags.

JRuby version      | Flags       | Method#arity   | Method#parameters
-------------------|-------------|----------------|------------------
1.7.0 - 1.7.27     | --1.8       | Java proxy bug | Not supported
1.7.0 - 1.7.19     | --1.9,--2.0 | Java proxy bug | Java proxy bug, attr_writer bug
1.7.20 - 1.7.27    | --1.9,--2.0 | Java proxy bug | Java proxy bug
9.0.0.0 - 9.0.1.0  |             | Java proxy bug | Java proxy bug
9.0.3.0 - 9.1.15.0 |             |                | Java proxy bug

Given that I was on JRuby 9.1.13.0, which has a working `Method#arity` implementation, I first implemented a workaround for that, but when running the tests I found out that `Method#arity` wouldn't help in a bunch of cases, but given that the #arity workaround would have to work in 1.8 mode (without parameters), I decided to implement that workaround completely separately. In order to clarify that, I split the two workarounds into different commits with better explanations.

lib/rspec/support/method_signature_verifier.rb
grddev: Fixed

lib/rspec/support/method_signature_verifier.rb
grddev: The motivation for having them separate is that the latter one cannot use `#parameters`, but it could obviously be implemented by checking support for `#parameters` at runtime, if you prefer such a solution.

lib/rspec/support/method_signature_verifier.rb
grddev: As should hopefully be clear by the table in my overall comment, there are no versions where the `attr_writer` bug exists, but the Java proxy method one does not.

lib/rspec/support/method_signature_verifier.rb
grddev: Fixed

lib/rspec/support/method_signature_verifier.rb
grddev: I explained it the following way in the new commit message, but would you prefer a comment in the code?
> The workaround essentially makes use of Java's introspection to figure out matching methods (which could be more than one partly because Java supports multiple overloads, and partly because JRuby introduces aliases to make method names look more Rubyesque). If there is only a single match, we can use that methods arity directly instead of the default -1 arity.

myronmarston:
Thanks @grddev!

lib/rspec/support/method_signature_verifier.rb
```diff
@@ -137,36 +149,46 @@ def unlimited_args?
...
+      end
+
+      # JRuby used to always report -1 arity for Java proxy methods
+      if Java::JavaLang::String.instance_method(:char_at).arity == -1
+        class MethodSignature < remove_const(:MethodSignature)
+        private
+
+          def classify_parameters
+            super
+            return unless @method.arity == -1 && @method.owner.respond_to?(:java_class)
```
JonRowe: I believe this is breaking the build for rspec/rspec-expectations, see https://travis-ci.org/rspec/rspec-expectations/jobs/351862257

grddev: Ah, I see now that the method is supposed to work for Proc as well as Method, and `#owner` isn't available for Proc. The easiest way to handle this would simply to add `@method.respond_to?(:owner)` to the referenced early return. This will effectively disable the workaround for Proc, while retaining it for Method.

While the same problem exists for Proc, and you could observe it using the following construction,
```
Java::JavaLang::String.new.method(:char_at).to_proc.arity # => -1
```
I can't think of a workaround that would address this case, as I'm not aware of any JRuby API that allows you to figure out if a Proc is a direct wrapper around a Java method.

JonRowe: Could we add the proc fix for now, and add a pending spec demonstrating the problem for the scenario described, this will document our knowledge on this

grddev: Opened a PR with the proposed fix in #347

PR #347, opened 2018-03-12
Adapt JRuby arity workaround to BlockSignature

grddev:
The workaround looking up the corresponding Java class and looking at its methods only works for methods, and doesn't work for Proc, where there isn't a direct API to access the proxied Java class.

While the same problem exists for Proc's, until https://github.com/jruby/jruby/issues/2817 has been resolved, there isn't much that can be done for that case. However, the current code fails to run for all blocks with arity -1, so this tweak is necessary to ensure normal blocks work again.

I couldn't find any specs for BlockSignature, so wasn't sure where it would be appropriate to add a test for this.


myronmarston:
Thanks @grddev!

PR #353, opened 2018-10-09
Prevent NoMethodError when setting a message expectation on an instance of a package-protected class

betesh:
An Java object can be an instance of a class that is
package-protected but implements or inherits from
a public type.  In that situation, the object's proxy
will only implement methods of its public interfaces
and superclasses, but `java_instance_methods` will
include methods that are defined on the
package-protected class.

For example:

    // Java code
    public interface A {
      public void runA();
    }

    class B implements A {
      public void runA() {}
      public void runB() {}
    }

    # Ruby code
    b = B.new
    b.java_class.java_instance_methods.map(&:name) #=> contains `:run_a` and `:run_b`
    b.method(:run_a) # => Returns a Method
    b.method(:run_b) # => raises a NoMethodError

### DIFF
--- a/lib/rspec/support/method_signature_verifier.rb
+++ b/lib/rspec/support/method_signature_verifier.rb
@@ -187,13 +187,19 @@ module RSpec
       if Java::JavaLang::String.instance_method(:char_at).arity == -1
         class MethodSignature < remove_const(:MethodSignature)
         private
+          def java_instance_methods
+            if @method.owner.java_class.public?
+              [@method.owner.java_class]
+            else
+              @method.owner.java_class.interfaces + [@method.owner.java_class.generic_superclass]
+            end.select(&:public?).map(&:java_instance_methods).flatten
+          end
 
           def classify_parameters
             super
             return unless @method.arity == -1
             return unless @method.respond_to?(:owner)
             return unless @method.owner.respond_to?(:java_class)
-            java_instance_methods = @method.owner.java_class.java_instance_methods
             compatible_overloads = java_instance_methods.select do |java_method|
               @method == @method.owner.instance_method(java_method.name)
             end


benoittgt:
Hello

Thanks for you PR. Do you think we could have a test to validate your patch?

lib/rspec/support/method_signature_verifier.rb
```diff
@@ -187,13 +187,19 @@ def classify_parameters
       if Java::JavaLang::String.instance_method(:char_at).arity == -1
         class MethodSignature < remove_const(:MethodSignature)
         private
+          def java_instance_methods
+            if @method.owner.java_class.public?
+              [@method.owner.java_class]
+            else
+              @method.owner.java_class.interfaces + [@method.owner.java_class.generic_superclass]
+            end.select(&:public?).map(&:java_instance_methods).flatten
+          end
```
JonRowe: Could we refactor this back inline below?
```
java_instance_methods =
  if @method.owner.java_class.public?
    @method.owner.java_class.java_instance_methods
  else
    (
      @method.owner.java_class.interfaces +
      [@method.owner.java_class.generic_superclass]
    ).inject([]) do |methods, klass|
      return methods unless klass.public? 
      methods += klass.java_instance_methods
  end
```

JonRowe:
Thanks @betesh this is an interesting bug you've found! You might wish to report it upstream to JRuby, can you add a comment explaining the work around, and could you add a test, preferably using standard java classes if possible.

lib/rspec/support/method_signature_verifier.rb
betesh: rubocop will complain about ABC size if I put this inline.  What do you suggest?

JonRowe: Hm ok, maybe refactor this as a seperate method then, I'd like to see an early return without any looping if the class is public

betesh:
I've been working on a test for this and I'm unable to reproduce it in rspec-support tests.  So it may be a bug in my application.  Closing for now.  I'll re-open this if I figure out how to reproduce it outside my application.

PR #366, opened 2019-02-25
Recognize hash as optional arg when optional keyword is present

edzhelyov:
I hit an issue in one of the projects that I work that I've reported in [rspec-mocks#1266](https://github.com/rspec/rspec-mocks/issues/1266).

In short, given `foo(a, b = {}, x: false)` should recognize a valid method call with args: `foo(1, a: 1, b: 2)`.

I've opted to extend the spec for optional keywords. I'm feeling lost in the specs, so please, give me some pointers if you need the specs re-worked.

I'm OK if you want to change the implementation or try to catch edge cases involving splats.

### DIFF
--- a/lib/rspec/support/method_signature_verifier.rb
+++ b/lib/rspec/support/method_signature_verifier.rb
@@ -77,17 +77,18 @@ module RSpec
           given_kw_args - @allowed_kw_args
         end
 
+        # If the last argument is Hash, Ruby will treat only symbol keys as keyword arguments
+        # the rest will be grouped in another Hash and passed as positional argument.
         def has_kw_args_in?(args)
-          Hash === args.last && could_contain_kw_args?(args)
+          Hash === args.last &&
+            could_contain_kw_args?(args) &&
+            args.last.keys.any? { |x| x.is_a?(Symbol) }
         end
 
         # Without considering what the last arg is, could it
         # contain keyword arguments?
         def could_contain_kw_args?(args)
           return false if args.count <= min_non_kw_args
-          return false if args.count <= max_non_kw_args &&
-            Hash === args.last &&
-            args.last.keys.none? { |x| x.is_a?(Symbol) }
 
           @allows_any_kw_args || @allowed_kw_args.any?
         end
--- a/lib/rspec/support/method_signature_verifier.rb
+++ b/lib/rspec/support/method_signature_verifier.rb
@@ -361,7 +361,14 @@ module RSpec
 
       def split_args(*args)
         kw_args = if @signature.has_kw_args_in?(args)
-                    args.pop.keys
+                    last = args.pop
+                    non_kw_args = last.reject { |k, _| k.is_a?(Symbol) }
+                    if non_kw_args.empty?
+                      last.keys
+                    else
+                      args << non_kw_args
+                      last.select { |k, _| k.is_a?(Symbol) }.keys
+                    end
                   else
                     []
                   end
--- a/lib/rspec/support/method_signature_verifier.rb
+++ b/lib/rspec/support/method_signature_verifier.rb
@@ -85,6 +85,10 @@ module RSpec
         # contain keyword arguments?
         def could_contain_kw_args?(args)
           return false if args.count <= min_non_kw_args
+          return false if args.count <= max_non_kw_args &&
+            Hash === args.last &&
+            args.last.keys.none? { |x| x.is_a?(Symbol) }
+
           @allows_any_kw_args || @allowed_kw_args.any?
         end
 
--- a/spec/rspec/support/method_signature_verifier_spec.rb
+++ b/spec/rspec/support/method_signature_verifier_spec.rb
@@ -366,6 +366,98 @@ module RSpec
           end
         end
 
+        if RubyFeatures.kw_args_supported?
+          describe 'a method with optional argument and keyword arguments' do
+            eval <<-RUBY
+              def arity_kw(x, y = {}, z:2); end
+            RUBY
+
+            let(:test_method) { method(:arity_kw) }
+
+            it 'does not require any of the arguments' do
+              expect(valid?(nil)).to eq(true)
+              expect(valid?(nil, nil)).to eq(true)
+            end
+
+            it 'does not allow an invalid keyword arguments' do
+              expect(valid?(nil, nil, :a => 1)).to eq(false)
+              expect(valid?(nil, :a => 1)).to eq(false)
+            end
+
+            it 'allows Hash containing strings as last argument' do
+              expect(valid?(nil, 'a' => 1)).to eq(true)
+            end
+
+            it 'mentions the invalid keyword args in the error', :pending => RSpec::Support::Ruby.jruby? && !RSpec::Support::Ruby.jruby_9000? do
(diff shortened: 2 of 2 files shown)


JonRowe:
Great start at fixing [rspec-mocks#1266](https://github.com/rspec/rspec-mocks/issues/1266) I started trying to write a spec for it but got distracted. 

What I would like to see here though is this being seperate cases demonstrating it in addition to our existing examples, as I want to ensure we don't regress other paths

edzhelyov:
Thanks, I will rework the specs.

lib/rspec/support/method_signature_verifier.rb
```diff
@@ -85,6 +85,10 @@ def has_kw_args_in?(args)
         # contain keyword arguments?
         def could_contain_kw_args?(args)
           return false if args.count <= min_non_kw_args
+          if args.count <= max_non_kw_args && Hash === args.last
```
edzhelyov: The reason I've added `Hash === args.last` is that we use `could_contain_kw_args?` in https://github.com/rspec/rspec-support/blob/master/lib/rspec/support/method_signature_verifier.rb#L383 and I'm guarding against the matcher.

spec/rspec/support/method_signature_verifier_spec.rb
```diff
@@ -366,6 +366,105 @@ def arity_kw(x, y:1, z:2); end
...
+              expect(valid?(nil, 'a' => 1, :b => 2, :z => 3)).to eq(false)
+            end
+
+            it 'mentions the invalid keyword args in the error', :pending => RSpec::Support::Ruby.jruby? && !RSpec::Support::Ruby.jruby_9000? do
+              expect(error_for(nil, nil, :a => 0)).to \
+                eq("Invalid keyword arguments provided: a")
+              expect(error_for(nil, :a => 0)).to \
+                eq("Invalid keyword arguments provided: a")
+              expect(error_for(nil, 'a' => 0, :b => 0)).to \
+                eq("Invalid keyword arguments provided: b")
```
JonRowe: Stylistically we prefer having these all on one line unless it hits the limits, then we'd divy it up like:
```
expect(
  ...
).to ...
```

lib/rspec/support/method_signature_verifier.rb
JonRowe: Fair enough, this could be inlined into one if though?

lib/rspec/support/method_signature_verifier.rb
```diff
@@ -85,6 +85,10 @@ def has_kw_args_in?(args)
         # contain keyword arguments?
```
edzhelyov: Not sure if this comment is relevant now after my changes...

spec/rspec/support/method_signature_verifier_spec.rb
```diff
@@ -366,6 +366,102 @@ def arity_kw(x, y:1, z:2); end
...
+              expect(valid?(nil, 'a' => 1)).to eq(true)
+              expect(valid?(nil, 'a' => 1, :z => 3)).to eq(true)
+              expect(valid?(nil, 'a' => 1, :b => 3)).to eq(false)
+              expect(valid?(nil, 'a' => 1, :b => 2, :z => 3)).to eq(false)
+            end
+
+            it 'mentions the invalid keyword args in the error', :pending => RSpec::Support::Ruby.jruby? && !RSpec::Support::Ruby.jruby_9000? do
+              expect(error_for(nil, nil, :a => 0)).to eq("Invalid keyword arguments provided: a")
+              expect(error_for(nil, :a => 0)).to eq("Invalid keyword arguments provided: a")
+              expect(error_for(nil, 'a' => 0, b: 0)).to eq("Invalid keyword arguments provided: b")
```
JonRowe: Need to `eval` this for 1.8.7 to be able to read the file

lib/rspec/support/method_signature_verifier.rb
JonRowe: It is, but is now confusing, its saying that could args contain keyword arguments without us checking what the last argument is specifically (to allow matches), thats still true except when theres not enough arguments, then we check for the hash in the middle so to speak

spec/rspec/support/method_signature_verifier_spec.rb
edzhelyov: @JonRowe ah, sorry. Everything is passing now.

edzhelyov:
@JonRowe, not sure if you have received notification, everything is passing now.

lib/rspec/support/method_signature_verifier.rb
JonRowe: Did you want to update the comment?

lib/rspec/support/method_signature_verifier.rb
```diff
@@ -85,6 +85,10 @@ def has_kw_args_in?(args)
...
         def could_contain_kw_args?(args)
           return false if args.count <= min_non_kw_args
+          return false if args.count <= max_non_kw_args &&
+            Hash === args.last &&
+            args.last.keys.none? { |x| x.is_a?(Symbol) }
```
JonRowe: Did this exceed the line length/

lib/rspec/support/method_signature_verifier.rb
```diff
@@ -357,7 +361,14 @@ def unlimited_args?
...
       def split_args(*args)
         kw_args = if @signature.has_kw_args_in?(args)
-                    args.pop.keys
+                    last = args.pop
+                    non_kw_args = last.reject { |k, _| k.is_a?(Symbol) }
+                    if non_kw_args.empty?
+                      last.keys
+                    else
+                      args << non_kw_args
+                      last.select { |k, _| k.is_a?(Symbol) }.keys
```
JonRowe: Are you saying with this that Ruby will mix keyword args and hash keys?

JonRowe:
Sorry for the delay @edzhelyov my cognitive load has been a bit too high to deal with this until now

lib/rspec/support/method_signature_verifier.rb
edzhelyov: Yes, 125 characters in total.

lib/rspec/support/method_signature_verifier.rb
edzhelyov: If you provide an anonymous Hash as the last argument only the symbols from that argument will be treated as keyword arguments, the rest will be interpreted as positional one.

```
def foo(x = {}, y:); puts "#{x}, #{y}"; end

foo 'a' => 1, y: 2
=> {"a"=>1}, 2
foo x: 3, y: 2
=> ArgumentError (unknown keyword: x)

def bar(x:, y:); end

bar 'x' => 1
=> ArgumentError (wrong number of arguments (given 1, expected 0; required keywords: x, y))
bar x: 1, y: 2, z: 3
=> ArgumentError (unknown keyword: z)
```

I'm trying to cover the behavior in [this spec](https://github.com/rspec/rspec-support/blob/778f143da167ed4f3f8e8715543ddec153694aaa/spec/rspec/support/method_signature_verifier_spec.rb#L387)

JonRowe:
Thanks so much for your hard work!

Issue #374, opened 2019-04-20
Master branch miscalculates the expected argument count when a hash is passed to rest of keywords


twalpole:
### Subject of the issue

Given an instance method on a class defined as

    def test(arg, **options) ... end

when creating a verified partial double from an instance of that class that calling it

   allow(some_object).to receive(:test)
   some_object.test('something', {})

the error `Wrong number of arguments. Expected 1, got 2` is returned which is incorrect since the empty hash is accepted by the rest of keywords `**options`

JonRowe:
Fixed in #374

PR #375, opened 2019-04-21
Empty last hash matches allow any kwargs

twalpole:
This fixes issue #374

### DIFF
--- a/lib/rspec/support/method_signature_verifier.rb
+++ b/lib/rspec/support/method_signature_verifier.rb
@@ -82,7 +82,7 @@ module RSpec
         def has_kw_args_in?(args)
           Hash === args.last &&
             could_contain_kw_args?(args) &&
-            args.last.keys.any? { |x| x.is_a?(Symbol) }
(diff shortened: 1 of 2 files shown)


JonRowe:
Ok, you'll need a spec for this, also consider not everything responds to `empty?`...

twalpole:
@JonRowe At that point args.last is guaranteed to be a Hash so it is guaranteed to respond to `empty?` and there is a spec already in the commit.

twalpole:
@JonRowe - Just a friendly ping

benoittgt:
Sorry for the late review. This patch looks good to me. Because we do `Hash === args.last` as the first condition we are sure `arg.last.empty?` is done on the proper type. Also @JonRowe I don't understand which test we should add.

For the failing on appveyor. It is unrelated. https://ci.appveyor.com/project/rspec/rspec-support/builds/23996036/job/egb1o129jo145r0t

JonRowe:
Thanks :)

PR #378, opened 2019-06-11
\r should be escaped when inside an array

JonRowe:
To prevent terminal rewinding, we should escape `\r`, replaces #293


benoittgt:
First failing on CI was related to JRuby and failure with rvm. I reran the three jobs. 💚

PR #392, opened 2019-12-26
Ruby 2.7 support

JonRowe:

### DIFF
--- a/lib/rspec/support/encoded_string.rb
+++ b/lib/rspec/support/encoded_string.rb
@@ -112,10 +112,39 @@ module RSpec
           string = remove_invalid_bytes(string)
           string.encode(@encoding)
         rescue Encoding::UndefinedConversionError, Encoding::InvalidByteSequenceError
-          string.encode(@encoding, ENCODE_UNCONVERTABLE_BYTES)
+          encode_unconvertable_bytes(string)
         rescue Encoding::ConverterNotFoundError
-          string.dup.force_encoding(@encoding).encode(ENCODE_NO_CONVERTER)
+          encode_no_converter(string.dup.force_encoding(@encoding))
         end
+
+        private
+
+        # On Ruby 2.7.0 keyword arguments mixed with conventional cause a warning to
+        # be issued requiring us to be explicit by using a ** to pass the hash as
+        # keyword arguments. Any keyword argument supporting Ruby supports this.
+        if RubyFeatures.kw_args_supported?
+          # Note on non keyword supporting Ruby ** causes a syntax error hence
+          # we must use eval. To be removed in RSpec 4.
+          binding.eval(<<-CODE, __FILE__, __LINE__)
+          def encode_unconvertable_bytes(string)
+            string.encode(@encoding, **ENCODE_UNCONVERTABLE_BYTES)
+          end
+
+          def encode_no_converter(string)
+            string.encode(**ENCODE_NO_CONVERTER)
+          end
+          CODE
+        else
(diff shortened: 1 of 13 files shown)


JonRowe:
The major change here is removing the warning by using `**kwargs` which is not universally compatible, hence the eval (ick) defined methods, I'm open to better solutions here but we are calling keyword arguments and there is no every Ruby parseable way to define this I know of.

lib/rspec/support/spec/stderr_splitter.rb
```diff
@@ -38,6 +38,10 @@ def to_io
       def write(line)
         return if line =~ %r{^\S+/gems/\S+:\d+: warning:} # http://rubular.com/r/kqeUIZOfPG
 
+        # Ruby 2.7.0 complains about hashes used in place of keyword arguments
+        # Aruba 0.14.2 uses this internally triggering that here
+        return if line =~ %r{lib/ruby/2\.7\.0/fileutils\.rb:622: warning:}
```
pirj: @mvz you might be interested

lib/rspec/support/encoded_string.rb
```diff
@@ -112,11 +112,40 @@ def matching_encoding(string)
...
 
+        private
+
+        # On Ruby 2.7.0 keyword arguments mixed with conventional cause a warning to
+        # be issued requiring us to be explicit by using a ** to pass the hash as
+        # keyword arguments. Any keyword argument supporting Ruby supports this.
+        if RubyFeatures.kw_args_supported?
+          # Note on non keyword supporting Ruby ** causes a syntax error hence
+          # we must use eval. To be removed in RSpec 4.
+          binding.eval(<<-CODE, __FILE__, __LINE__)
```
benoittgt: Thanks Jon for the patch.
That's sad we have to do that to deal with Ruby 2.7.

JonRowe: Yeah I can't think of another away around it except having an extra file thats not required by the older build... I'm unsure if I like that less...

benoittgt: Yes me too. The comments help for this part. 

Thanks

lib/rspec/support/spec/stderr_splitter.rb
mvz: Thanks @pirj. This is fixed in Aruba 0.14.14.

lib/rspec/support/encoded_string.rb
mvz: While fixing 2.7 compatibility for aruba, I found you can still pass the keyword arguments as a 1.8-style hash literal. This would allow getting rid of the eval. You would also lose the constants though.

JonRowe: @mvz I think that depends on the API in use, I tried the hash directly with `String#encode` and it still produces the warning.

mvz: @JonRowe I'm not sure how that happened, but I just had to try as well and it seems to work here too: https://github.com/rspec/rspec-support/pull/394.

JonRowe: 🤦‍♂ 
```Ruby
2.7.0 > "abc".encode("UTF-8", {:invalid => :replace})
(irb):2: warning: Using the last argument as keyword parameters is deprecated
 => "abc"
2.7.0 > "abc".encode("UTF-8", :invalid => :replace)
 => "abc"
```

pirj: 🤦‍♂

PR #394, opened 2019-12-29
Use old-style hashes to pass keyword arguments

mvz:


JonRowe:
Would you mind rebasing to remove my merge commit?

lib/rspec/support/encoded_string.rb
```diff
@@ -112,34 +112,15 @@ def matching_encoding(string)
...
-          CODE
-        else
-          def encode_unconvertable_bytes(string)
-            string.encode(@encoding, ENCODE_UNCONVERTABLE_BYTES)
-          end
-
-          def encode_no_converter(string)
-            string.encode(ENCODE_NO_CONVERTER)
-          end
+          # ENCODE_NO_CONVERTER inlined here to avoid warnings on Ruby 2.7
```
pirj: Nice! 👏 
Do you think it's ok to drop those constants?

JonRowe: Ah yep, but the docs then needed sorting so, sorted :joy:

PR #395, opened 2020-01-10
Update ripper_supported? for truffleruby and ripper specs

bjfish:
and allow ripper specs to produce exceptions.

### DIFF
--- a/lib/rspec/support/ruby_features.rb
+++ b/lib/rspec/support/ruby_features.rb
@@ -47,6 +47,10 @@ module RSpec
       def mri?
         !defined?(RUBY_ENGINE) || RUBY_ENGINE == 'ruby'
       end
+
+      def truffleruby?
+        defined?(RUBY_ENGINE) && RUBY_ENGINE == 'truffleruby'
+      end
     end
 
     # @api private
@@ -101,7 +105,7 @@ module RSpec
       end
       ripper_requirements = [ComparableVersion.new(RUBY_VERSION) >= '1.9.2']
 
-      ripper_requirements.push(false) if Ruby.rbx?
+      ripper_requirements.push(false) if Ruby.rbx? || Ruby.truffleruby?
 
       if Ruby.jruby?
         ripper_requirements.push(Ruby.jruby_version >= '1.7.5')
--- a/spec/rspec/support/ruby_features_spec.rb
+++ b/spec/rspec/support/ruby_features_spec.rb
@@ -111,7 +111,7 @@ module RSpec
           in_sub_process_if_possible do
             begin
               require 'ripper'
-              !!defined?(::Ripper)
+              !!defined?(::Ripper) && Ripper.respond_to?(:lex)
             rescue LoadError
               false
             end


lib/rspec/support/ruby_features.rb
```diff
@@ -101,7 +105,7 @@ def supports_taint?
       end
       ripper_requirements = [ComparableVersion.new(RUBY_VERSION) >= '1.9.2']
 
-      ripper_requirements.push(false) if Ruby.rbx?
+      ripper_requirements.push(false) if Ruby.rbx? || Ruby.truffleruby?
```
eregon: This is not ideal, because TruffleRuby might support `Ripper` at some point.
How about checking `Ripper.respond_to?(:lex)` which is what https://github.com/k0kubun/hamlit does for instance?

JonRowe: You could push that onto the stack for if run truffle ruby?

bjfish: @eregon I think there would be an issue with this because part of the spec asserts `Ripper` would not be loaded by feature detection: https://github.com/rspec/rspec-support/blob/b83b4cecb3bb56300addae89cb0c437f1af906fb/spec/rspec/support/ruby_features_spec.rb#L151-L153

spec/rspec/support/ruby_features_spec.rb
```diff
@@ -126,21 +126,29 @@ def ripper_works_correctly?
...
+            begin
+              require 'ripper'
+              tokens = ::Ripper.lex('foo')
+              token = tokens.first
+              location = token.first
+              line_number = location.first
+              line_number == 1
+            rescue
+              false
+            end
```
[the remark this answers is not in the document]
bjfish: I've updated the PR to not run these sections if `RubyFeatures.ripper_supported?` is `false`.

spec/rspec/support/ruby_features_spec.rb
```diff
@@ -126,21 +126,29 @@ def ripper_works_correctly?
...
-            require 'ripper'
-            sexp = ::Ripper.sexp(':if')
-            !sexp.nil?
+            begin
+              require 'ripper'
+              sexp = ::Ripper.sexp(':if')
+              !sexp.nil?
+            rescue
+              false
+            end
```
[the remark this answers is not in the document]
bjfish: See https://github.com/rspec/rspec-support/pull/395#discussion_r365466032

lib/rspec/support/ruby_features.rb
eregon: I see, that is unfortunate. Probably fine to keep as-is then and change it again (e.g., to a version check) once TruffleRuby supports Ripper.

spec/rspec/support/ruby_features_spec.rb
```diff
@@ -145,7 +145,7 @@ def ripper_can_parse_source_including_keywordish_symbol?
         end
 
         it 'returns whether Ripper is correctly implemented in the current environment' do
-          expect(RubyFeatures.ripper_supported?).to eq(ripper_is_implemented? && ripper_works_correctly?)
+          expect(RubyFeatures.ripper_supported?).to eq(RubyFeatures.ripper_supported? && ripper_is_implemented? && ripper_works_correctly?)
```
pirj:
```ruby
expect(RubyFeatures.ripper_supported?).to eq(RubyFeatures.ripper_supported?
```
That looks odd quite frankly :D

JonRowe: This change should be reverted, as `ripper_supported?` will be `false`, the test methods should return false as ripper isn't implemented properly on truffle right?

lib/rspec/support/ruby_features.rb
JonRowe:
```suggestion
      ripper_requirements.push(defined?(::Ripper) && Ripper.respond_to?(:lex)) if Ruby.truffleruby?
```

spec/rspec/support/ruby_features_spec.rb
JonRowe: This should be reverted, the check here is that `ripper_supported?` matches the two test definitions, are you saying the test definitions return `true` here? As that should be impossible if ripper is broken on truffleruby. Whats actually happening?

bjfish: @JonRowe On `truffleruby`, a runtime error occurs when running `ripper_is_implemented? && ripper_works_correctly?` which is why I suggested the `begin/rescue` change before to assert that running these results in error when `RubyFeatures.ripper_supported?` is `false`.

Per your comment: https://github.com/rspec/rspec-support/pull/395#discussion_r365361330. I updated to not run this spec if `RubyFeatures.ripper_supported?` is `false`.

@pirj I agree this looked odd. I've pushed a new version to be more explicit in excluding this spec from running using: `, :if => ripper_supported?`. Please review.

spec/rspec/support/ruby_features_spec.rb
```diff
@@ -123,6 +123,10 @@ def ripper_works_correctly?
             ripper_can_parse_source_including_keywordish_symbol?
         end
 
+        def self.ripper_supported?
+          RubyFeatures.ripper_supported?
```
JonRowe: This is the subject under test, its not a valid way to see if ripper is supported, `defined?(Ripper)` would be ok.

JonRowe:
> This should be reverted, the check here is that ripper_supported? matches the two test definitions, are you saying the test definitions return true here? As that should be impossible if ripper is broken on truffleruby. Whats actually happening?

Github has lost the reply to this somewhere, whats the runtime error from those two detection methods? Is that what the `rescue false` was for originally? Out of those two solutions I'd prefer you rescue the specific exception with a note about truffle-ruby.

bjfish:
@JonRowe I've updated to add the `begin;rescue` with a note for truffleruby and removed the dependency on `RubyFeatures.ripper_supported?` to run the test. Thanks.

lib/rspec/support/ruby_features.rb
eregon: That would not work because to access the Ripper constant one would need to `require 'ripper'` but there is a spec checking that as @bjfish wrote above.

eregon: Although that spec would likely not see whether `ripper` was required before `ripper_supported?` is defined. But I guess the intent is that file doesn't load ripper.

spec/rspec/support/ruby_features_spec.rb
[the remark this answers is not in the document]
JonRowe: Ah ok, so this comes from the `lex`? implementation? Looking at the code a better fix would to change the check in `is_implemented?` expanding it to `!!defined?(::Ripper) && Ripper.respond_to?(:lex)`.

bjfish:
@JonRowe I've updated `ripper_is_implemented?` with your suggestion, thanks.

lib/rspec/support/ruby_features.rb
JonRowe: @eregon `defined?(::Ripper)` would return false unless you've required it, which is the same as this behaviour, but then if it has been required, it will check for the `lex` method to see if truffle ruby is now compatible with ripper support.

eregon: How does this work in practice in RSpec? If it wants to use Ripper, it will have to require it, no?
Or does it rely on the user requiring Ripper to use it?

JonRowe: > How does this work in practice in RSpec? If RSpec wants to use Ripper, it will have to require it, no? Or does it rely on the user requiring Ripper to use it?

We require it if ripper is supported, but on recent MRI it's already loaded. In practise it would mean that you'd have to have require it on truffle ruby if you wanted to use it and it would work.

eregon: > We require it if ripper is supported,

Right, so the whole point of this logic is then avoiding `require 'ripper'` if Ripper is unsupported.
Is that to avoid changing `$LOADED_FEATURES` or making sure code which uses Ripper still `require 'ripper'`, but only for Rubies with unsupported Ripper?

I would suggest something a lot more straightforward by always requiring Ripper, then we wouldn't need to have so much Ruby-implementation-specific code here.

> on recent MRI it's already loaded

I don't see that:
```
$ ruby -ve 'puts $".grep(/ripper/i)'
ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-linux]
```

> In practise it would mean that you'd have to have require it on truffle ruby if you wanted to use it and it would work.

I wouldn't want that, I'd want the same behavior as MRI.
Since anyway this does not seem a critical feature of RSpec, I propose to adapt the code here once TruffleRuby supports Ripper and merge the PR as it is.

JonRowe: >Right, so the whole point of this logic is then avoiding require 'ripper' if Ripper is unsupported. Is that to avoid changing $LOADED_FEATURES or making sure code which uses Ripper still require 'ripper', but only for Rubies with unsupported Ripper?

It's to avoid requiring ripper if possible.

> I would suggest something a lot more straightforward by always requiring Ripper, then we wouldn't need to have so much Ruby-implementation-specific code here.

We do the least requiring we can to avoid false positives due to poisoned environments.

> I don't see that:

I stand corrected, `irb` requires it, `ruby` does not.

> I wouldn't want that, I'd want the same behavior as MRI.

In my experience only MRI behaves like MRI, other Ruby engines always have quirks. 😢

eregon: > In my experience only MRI behaves like MRI, other Ruby engines always have quirks.

FWIW, TruffleRuby tries hard to behave exactly like MRI. But implementing Ripper is a significant effort, so that's not done yet.

JonRowe: To be fair I should probably say that every engine has quirks, it's just the MRI ones are the reference quirks? 🤷‍♂

eregon: I just wanted to end on a positive note :)
This PR is not the most beautiful code but TruffleRuby otherwise runs RSpec (and Rails nowadays) unmodified with no TruffleRuby-specific code since many years now.
We try hard to avoid any TruffleRuby-specific code in gems.

PR #396, opened 2020-01-10
Prevent Rakefile from loading rubocop when dependency is not present

bjfish:
I encountered the following error:
```
% bundle exec rake spec
rake aborted!
cannot load such file -- rubocop/rake_task
```
because the `Gemfile` has the following condition for including this dependency:
```
if RUBY_VERSION >= '2.4' && RUBY_ENGINE == 'ruby'
  gem "rubocop", "~> 0.52.1"
end
```
So, I've added the same condition to the `Rakefile`.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -23,8 +23,12 @@ end
 
 task :build => :verify_private_key_present
 
-require 'rubocop/rake_task'
-desc 'Run RuboCop on the lib directory'
-RuboCop::RakeTask.new(:rubocop) do |task|
-  task.patterns = ['lib/**/*.rb']
+begin
+  require 'rubocop/rake_task'
+  desc 'Run RuboCop on the lib directory'
+  RuboCop::RakeTask.new(:rubocop) do |task|
+    task.patterns = ['lib/**/*.rb']
+  end
+rescue LoadError
+  # No rubocop means no rubocop rake task
 end


Rakefile
```diff
@@ -23,8 +23,10 @@ end
 
 task :build => :verify_private_key_present
 
-require 'rubocop/rake_task'
-desc 'Run RuboCop on the lib directory'
-RuboCop::RakeTask.new(:rubocop) do |task|
-  task.patterns = ['lib/**/*.rb']
+if RUBY_VERSION >= '2.4' && RUBY_ENGINE == 'ruby'
+  require 'rubocop/rake_task'
```
JonRowe: I'd prefer you just rescue the load error...

JonRowe: It's preferable to wrap this in a rescue block instead:
e.g.
```
begin
  ... 
rescue LoadError
  # No rubocop means no rubocop rake task
end
```

bjfish: I've updated to rescue the `LoadError`.

pirj:
It seems there's no consistency in how we define `rubocop` task across projects.

[`rspec-expectations`](https://github.com/rspec/rspec-expectations/blob/master/Rakefile#L12):
```ruby
if RUBY_VERSION >= '2.4' && RUBY_ENGINE == 'ruby'
  require 'rubocop/rake_task'
  RuboCop::RakeTask.new(:rubocop)
end

if RUBY_VERSION >= '2.4' && RUBY_ENGINE == 'ruby'
  task :default => [:spec, :cucumber, :rubocop]
else
```

[`rspec-core`](https://github.com/rspec/rspec-core/blob/master/Rakefile#L26):
```ruby
desc 'Run RuboCop on the lib directory'
task :rubocop do
  sh 'bundle exec rubocop lib'
end
```

Do you plan to touch those Rakefiles as well @bjfish ?

bjfish:
@pirj I'd probably leave them alone unless I had some issue. They appear they'd work as is for my use case:  `bundle exec rake spec`.

JonRowe:
@pirj I think it was done to limit what rubocop is applied to, but it also addresses this issue in a different way.

PR #401, opened 2020-03-21
Ignore Ruby 2.7 multiline warnings from gems

JonRowe:
This fixes the builds as its a Rails warning creeping through.

### DIFF
--- a/lib/rspec/support/spec/stderr_splitter.rb
+++ b/lib/rspec/support/spec/stderr_splitter.rb
@@ -6,6 +6,7 @@ module RSpec
       def initialize(original)
         @orig_stderr    = original
         @output_tracker = ::StringIO.new
+        @last_line = nil
       end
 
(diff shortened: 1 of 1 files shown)


lib/rspec/support/spec/stderr_splitter.rb
```diff
@@ -42,8 +43,14 @@ def write(line)
         # Aruba 0.14.2 uses this internally triggering that here
         return if line =~ %r{lib/ruby/2\.7\.0/fileutils\.rb:622: warning:}
 
+        # Ruby 2.7.0 warnings from keyword argments span multiple lines
```
pirj: Maybe put this closer together with the check for the first line of the warning?
```ruby
return if line =~ %r{^\S+/gems/\S+:\d+: warning:}
```
It took me a second to figure out how are we skipping the first line of it.

JonRowe: See change

benoittgt:
oooh thanks!

PR #407, opened 2020-04-06
Add helper for running class_exec with keywords when needed

JonRowe:
This is an extraction from a pr on rspec-core but its going to be a repeated pattern to fix keyword arguments across the gem so I'm adding it here. I expect to add more as I find other places that keyword arguments don't work quite right.

The aim will be to keep the string eval to a minimum and this will all be replaced with proper `**kwargs` in RSpec 4, which I want to release in the later half of this year before the next version of Ruby.

### DIFF
--- a/lib/rspec/support/with_keywords_when_needed.rb
+++ b/lib/rspec/support/with_keywords_when_needed.rb
@@ -1,0 +1,33 @@
+RSpec::Support.require_rspec_support("method_signature_verifier")
+
+module RSpec
+  module Support
+    module WithKeywordsWhenNeeded
+      # This module adds keyword sensitive support for core ruby methods
+      # where we cannot use `ruby2_keywords` directly.
+
+      module_function
+
+      if RSpec::Support::RubyFeatures.kw_args_supported?
+        # Remove this in RSpec 4 in favour of explictly passed in kwargs where
+        # this is used. Works around a warning in Ruby 2.7
+
+        def class_exec(klass, *args, &block)
+          if MethodSignature.new(block).has_kw_args_in?(args)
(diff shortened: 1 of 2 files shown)


lib/rspec/support/with_keywords_when_needed.rb
```diff
@@ -0,0 +1,33 @@
+RSpec::Support.require_rspec_support("method_signature")
```
benoittgt, line 1: Isn't `method_signature_verifier` that you want to require?

JonRowe: It is indeed, pushed up for the expectations build without checking :joy:

JonRowe:
Merged because something odd is going on with the sub builds which means this wasn't picked up on the pinned branch.

spec/rspec/support/with_keywords_when_needed_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      def run(klass, *args, &block)
+        WithKeywordsWhenNeeded.class_exec(klass, *args, &block)
+      end
+
+      it "will run a block without keyword arguments" do
+        run(klass, 42) { |arg| check_argument(arg) }
+      end
+
+      it "will run a block with a hash with out keyword arguments" do
```
pirj, line 25: "with out" => "without"

benoittgt:
On this subject https://discuss.rubyonrails.org/t/new-2-7-3-0-keyword-argument-pain-point/74980 (cc @JonRowe @pirj )

JonRowe:
I've added a reply

PR #411, opened 2020-04-26
Allow mocking Mutex.new

JonRowe:


JonRowe:
> I hope `new` is the only method that users would stub :D

Its the only one that matters / breaks, stubbing on an individual instance of mutex would affect only those instances, and stubbing using any_instance is just going to be something we don't support (its not really reasonable to).

PR #413, opened 2020-05-12
Add `WithKeywordsWhenNeeded.send_to` to handle keywords when using send with objects out of our control

JonRowe:


JonRowe:
I've not merged this yet because of the performance implications, its quite a lot slower, what do we feel about using more evaling of source code?

benoittgt:
Sorry Jon. I didn't see your question.

PR #421, opened 2020-06-27
`diff-lcs` has changed its format in 1.4

JonRowe:
The recent `diff-lcs` update changed some parts of the diff output, breaking our builds, these seem trivial enough to ignore (e.g. they don't seem wrong, just different).

Draft because I'm going to improve the formatting of this spec a bit whilst I'm here, and I'm going to setup CI to test both versions of `diff-lcs`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,12 @@ else
   gem 'rake', '>= 12.3.3'
 end
 
+if ENV['DIFF_LCS_VERSION']
+  gem 'diff-lcs', ENV['DIFF_LCS_VERSION']
+else
+  gem 'diff-lcs', '~> 1.4', '>= 1.4.3'
+end
+
 if RUBY_VERSION < '2.2.0' && !!(RbConfig::CONFIG['host_os'] =~ /cygwin|mswin|mingw|bccwin|wince|emx/)
   gem "childprocess", "< 1.0.0"
 elsif RUBY_VERSION < '2.3.0'
--- a/spec/rspec/support/differ_spec.rb
+++ b/spec/rspec/support/differ_spec.rb
@@ -7,16 +7,7 @@ require 'rspec/support/spec/string_matcher'
 module RSpec
   module Support
     RSpec.describe Differ do
-
-      # In the updated version of diff-lcs several diff headers change format slightly
-      # compensate for this and change minimum version in RSpec 4
-      if Diff::LCS::VERSION.to_f < 1.4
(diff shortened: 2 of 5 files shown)


JonRowe:
Related I've asked the diff-lcs team to bump their major version if they make Ruby changes in halostatue/diff-lcs#62

halostatue:
Can you detail the changes that were breaking for you on this with examples as a bug against `diff-lcs`? As I mentioned in halostatue/diff-lcs#62, I am willing to consider some changes to how diff-lcs works when running under RSpec so that we keep compatibility as much as possible.

JonRowe:
@halostatue This PR contains examples of the changed output, the line number headers have changed in some cases, and some unchanged lines have been removed, I'm totally fine with working around that in the two versions like this PR does :)

Gemfile
```diff
@@ -21,6 +21,8 @@ else
   gem 'rake', '>= 12.3.3'
 end
 
+gem 'diff-lcs', ENV.fetch('DIFF_LCS_VERSION', '~> 1.4')
```
benoittgt, line 24: Don't we want to stick to old flag like `if RUBY_VERSION`?

spec/rspec/support/differ_spec.rb
```diff
@@ -92,12 +135,13 @@ def differ_ivars
...
-
-@@ -1,2 +1,2 @@
--Tu avec carte {count} item has
-+Tu avec carté {count} itém has
-EOD
+            expected_diff = dedent(<<-EOD).encode('UTF-16LE')
+              |
+              |@@ #{one_line_header} @@
+              |-Tu avec carte {count} item has
+              |+Tu avec carté {count} itém has
```
benoittgt, line 142: Lol this means nothing in french. 😅

Gemfile
halostatue: If diff-lcs 1.x fails on < Ruby 2.5 it’s _still_ a failure, even though I’m no longer running CI on those versions.

spec/rspec/support/differ_spec.rb
JonRowe: Nothing like a good bit of franglais, its only the `é` thats important if you'd like to make it make sense :P

Gemfile
benoittgt: Oups sorry. I just saw "I'm going to setup CI to test both versions of diff-lcs."

PR #431, opened 2020-09-28
Allow File.read to be stubbed out without breaking ripper support

JonRowe:
Following on from conversation in rspec/rspec-mocks#1341 this fixes issues like rspec/rspec-support#428 when `File.read` is stubbed out and returns non standard content.


benoittgt (approved):
LGTM Jon

pirj (approved):
Awesome, this works nicely. Thanks!

PR #437, opened 2020-11-16
Updates from rspec-dev (2020-11-16)

JonRowe:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-# This file was generated on 2020-09-28T21:53:54+02:00 from the rspec-dev repo.
+# This file was generated on 2020-11-16T09:31:11+00:00 from the rspec-dev repo.
 # DO NOT modify it by hand as your changes will get lost the next time it is generated.
 
 # In order to install old Rubies, we need to use old Ubuntu distibution.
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -52,3 +43,4 @@ branches:
   only:
     - main
     - /^\d+-\d+-maintenance$/
+    - /^\d+-\d+-dev$/
```
pirj, line 46: :heart:

pirj (approved):
Is it possible to only run one build? Either for the branch (`main`, `*-maintenance`, `*-dev`) or for the PR?

JonRowe:
> Is it possible to only run one build? Either for the branch (main, *-maintenance, *-dev) or for the PR?

Don't know, my standard knowledge says that you need `pr` so that, er pull requests are built, but you also need `push` so that branches pushed up by us run.

JonRowe:
We might be able to limit push to `main` and maintenance branches?

JonRowe:
This sometimes happens with Travis as well by the way, it's just we don't see it

pirj:
> This sometimes happens with Travis as well by the way, it's just we don't see it

We found a reason for that with Benoit, it's that removed `branches` setting for `rspec-rails`. Doesn't happen with other repos.

JonRowe:
> We found a reason for that with Benoit, it's that removed branches setting for rspec-rails. Doesn't happen with other repos.
That branches option was deliberately limiting builds on rspec-rails becauase of the load it causes.

pirj:
> That branches option was deliberately limiting builds on rspec-rails 

`branches` option is [not only in `rspec-rails`](https://github.com/rspec/rspec-support/blob/9a2a668c83b858901052329518da414c299405ac/.travis.yml#L51):
```yaml
branches:
  only:
    - main
    - /^\d+-\d+-maintenance$/
```

Frankly, we have everywhere except `rspec-rails`, resulting in exactly the opposite effect :laughing:

PR #440, opened 2020-11-21
Updates from rspec-dev (2020-11-21)

JonRowe:


pirj:
> Ripper is supported on 9.1.17.0 according to test suite

👍 Good to know. I indeed to remove one more method from RubyFeatures in `drop-old-rubies`.

JonRowe:
`ripper_supported?` is not a method that should be removed.

pirj:
@JonRowe, do you mind to elaborate why, to support JRuby 9.1?

JonRowe:
Because we need it to flag ripper support on Rubies that don't.

pirj:
Right, Rubinius slipped my mind completely 👍
I had an impression that there's a `jruby?` check when defining `ripper_supported?` in `drop-old-rubies`, but `.rbx?` is the only one.

PR #442, opened 2020-11-22
Restore original JRuby ripper suppression

JonRowe:
A bunch of tests elsewhere fail when this is changed, so this branch is a reversal to double check that it fixes that.


JonRowe:
@pirj I'm merging this because it fixes JRuby 9.1.17.0 for rspec-expectations just as a heads up for your supported version work.

pirj:
> fixes JRuby 9.1.17.0 for rspec-expectations just as a heads up for your supported version work.

Thanks, noted 👍

PR #445, opened 2020-12-08
Updates from rspec-dev (2020-12-08)

JonRowe:


pirj:
It's beyond my understanding how this passes when `.travis.yml` is removed, but [this line is still in place](https://github.com/rspec/rspec-support/pull/445/files#diff-f686af268076eb6078295d47e8d97ab3df9486f0f22b900f4c045b5b9cc04fe7R91):
```
bundle_install_flags=`cat .travis.yml | grep bundler_args | tr -d '"' | grep -o " .*"`
```

pirj (approved):
Awesome, thanks! 💯

JonRowe:
> It's beyond my understanding how this passes when .travis.yml is removed, but this line is still in place:
> ```
> bundle_install_flags=`cat .travis.yml | grep bundler_args | tr -d '"' | grep -o " .*"`
> ```
 
It's empty when it fails, and bundle has already happened in github actions, so it presumedly has the config from that cached. This can be cleaned up at a later date.

JonRowe:
> This can be cleaned up at a later (post 4.0.0 merge to master) date.

Or applied to both I guess, just it need not be a 4.0.0 specific change

pirj:
> It's empty when it fails

I guess we don't set `set -e`, so this subshell call doesn't fail the caller if it fails itself, and we just go on.
👍

PR #446, opened 2020-12-08
Updates from rspec-dev (2020-12-08)

JonRowe:


pirj:
Do we need this in `main`? We're keeping .travis.yml around for older Ruby builds. Did you intend to target 4-0-dev?

JonRowe:
> Do we need this in main? We're keeping `.travis.yml` around for older Ruby builds. Did you intend to target `4-0-dev`?

The builds are breaking without it, apparently because the ci changes earlier ran with the non merged versions, and as I stated if we're changing this I want parity to reduce diffs and eventual review sizes.

JonRowe:
Travis build failure is due to older rubies breaking down there.

PR #475, opened 2020-12-29
Add support for `.call` to WithKeywordsWhenNeeded

pirj:
Needed for `rspec-mocks`

https://github.com/rspec/rspec-mocks/issues/1306

This is based on https://github.com/rspec/rspec-mocks/pull/1324/files#diff-afc11368b62c27a5bd155b766ec9cf4e96a9331d073b6123a9bce36a270de640R359

### DIFF
--- a/lib/rspec/support/with_keywords_when_needed.rb
+++ b/lib/rspec/support/with_keywords_when_needed.rb
@@ -19,6 +19,16 @@ module RSpec
         end
       end
       ruby2_keywords :class_exec if respond_to?(:ruby2_keywords, true)
+
+      def call(method, *args, &block)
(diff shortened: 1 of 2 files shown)


JonRowe:
This is the wrong approach, this class actually needs to be removed in RSpec 4. My approach using the helper class is the cause of the `to_hash` issues caused elsewhere, there is no way to determine between a hashlike object and keywords being passed through.

pirj:
`WithKeywordsWhenNeeded` bothers me as well.

I haven't managed to rewrite `rspec-core`'s `lib/rspec/core/shared_example_group.rb` with it.
Can you take the lead and get rid of it there? I can mimic the fix for `rspec-mocks` then.

PR #494, opened 2021-02-01
Verify partial doubles by default (dependent)

pirj:
See https://github.com/rspec/rspec-mocks/pull/1409

### DIFF
--- a/spec/rspec/support/recursive_const_methods_spec.rb
+++ b/spec/rspec/support/recursive_const_methods_spec.rb
@@ -12,6 +12,9 @@ module RSpec
 
         class Bar < Parent
           VAL = 10
+          def to_str
+            "Bar"
+          end
         end
       end
 


spec/rspec/support/recursive_const_methods_spec.rb
```diff
@@ -33,7 +36,6 @@ class Bar < Parent
         end
 
         it 'does not blow up on buggy classes that raise weird errors on `to_str`' do
-          allow(Foo::Bar).to receive(:to_str).and_raise("boom!")
```
JonRowe, line 36: I'd restore this and just add an implementation that doesn't raise? Seems like a "safer" change

pirj: 👍 Done

JonRowe (approved):

PR #502, opened 2021-04-25
Use `Fiber.current` for current execution context. Fixes #501.

### LINKED ISSUE #501: ThreadError: Attempt to unlock a mutex which is locked by another thread/fiber
### Subject of the issue

image

### Your environment
* Ruby version: 3.0
* rspec-support version: 3.10.2

### Steps to reproduce

This is a tricky issue to reproduce.

Essentially it looks like some of the assumptions made by `reentrant_mutex.rb` are no longer true in Ruby 3.0. I will review the code and try to give an update, but essentially it looks like trying to lock a `ReentrantMutex` from a different fiber on the same thread might be a problem.

ioquatix:


lib/rspec/support/reentrant_mutex.rb
```diff
@@ -28,8 +28,8 @@ def synchronize
     private
 
       def enter
-        @mutex.lock if @owner != Thread.current
-        @owner = Thread.current
```
pirj, line 32: Do we need to `require 'fiber'` somewhere? `Fiber.current` errors out in `irb -f`.

ioquatix: Oh yeah probably.

ioquatix:
@eregon wdyt?

ioquatix:
I'm not sure what the best option is to resolve the failure - do we add `fiber` to the list of exceptions?

lib/rspec/support/reentrant_mutex.rb
```diff
@@ -1,3 +1,5 @@
+require 'fiber'
```
pirj, line 1: I foresee that @JonRowe would say that requiring a dependency from stdlib is something that we try to avoid at all costs.

lib/rspec/support/reentrant_mutex.rb
```diff
@@ -28,8 +30,8 @@ def synchronize
     private
 
       def enter
-        @mutex.lock if @owner != Thread.current
-        @owner = Thread.current
+        @mutex.lock if @owner != Fiber.current
+        @owner = Fiber.current
```
pirj, line 34: WDYT of
```ruby
def current
  defined?(Fiber.current) ? Fiber.current : Thread.current
end
```
and using it?

pirj (commented):
We're dropping old rubies (< 2.3) in RSpec 4.
What are your primary targets? Do you still plan to support old rubies with pre-4.0 RSpec?

Does it make sense for RSpec to remove ReentrantMutex and use Monitor?
Or is it the case that Monitor has that same bug with `Fiber.yield` inside `synchronize`, and we'd better keep `ReentrantMutex` with what we'll end up doing in this PR that would work with any Ruby version? And the plan is to keep it until we drop support for Ruby 3.0?

pirj:
> the best option is to resolve the failure

My guess is that it compares loaded features to a clean process. It's a defensive measure to avoid loading anything explicitly, including `'fiber'`.

lib/rspec/support/reentrant_mutex.rb
ioquatix: It's broken.

lib/rspec/support/reentrant_mutex.rb
ioquatix: Fair enough.

ioquatix:
I would suggest that if monitor worked, you should use it, but it's broken by the same bug. Ideally we fix this in 3.0.2

eregon:
The change looks good to me.

Unfortunately there is no way to get `Fiber.current` without `require 'fiber'` in Ruby code (maybe we should change that in Ruby). `fiber` is not a default gem or anything you could replace, so requiring `fiber` seems fairly safe to me.
It does have the downside that a gem might appear to work when running specs, but not when used outside of specs, e.g., if it uses `Fiber.current`, `Fiber#alive?` or `Fiber#transfer` (fairly unlikely I'd say).
That's a general issue though, any test dependency might require stdlibs, and IMHO there should be at least one integration test using a subprocess to make sure this isn't an issue while loading the gem's files.

Using `monitor` seems best when available, but that's also a stdlib.
`monitor` is also not a default gem, since it is a dependency of RubyGems.
Because RubyGems already requires it, I think there is little problems to require it & use it in RSpec.

pirj:
@eregon We don't require Rubygems as well, and even explicitly avoid loading them on CI.
Rails, e.g. refers to Rubygems in their to securely compare gem versions, and don't explicitly require it that would break if Rails is run with Rubygems disabled with `rubyopt`. We deliberately don't require anything from stdlib unless absolutely necessary to prevent such breakages.

lib/rspec/support/reentrant_mutex.rb
pirj: Can you please elaborate?

eregon: Whenever some file would `require 'fiber'` the behavior and what's stored here would switch and it'd be a total mess.

eregon:
I think one way to avoid needing any stdlib here is to use `Mutex#owned?` and remove the `@owner` field entirely.
Then neither `Thread.current` or `Fiber.current` would be used, and it'd be correct.

lib/rspec/support/reentrant_mutex.rb
pirj: > some file would require 'fiber' the behavior and what's stored here would switch and it'd be a total mess

For sure. I hope they don't do it dynamically when there are existing fibers and reentrant mutexes in action. Otherwise - do you see any issue?

Ok, let's put it another way. I couldn't find any mention of `Fiber` in stdlib docs. [Core docs](https://ruby-doc.org/core-2.7.1/Fiber.html#method-c-current) claim that in order to user `current`/`transfer`/`alive?` one has to `require 'fiber'`.
Unlike it is with `require 'rubygems'` or `require 'date'`, `require 'set'` etc no constants are loaded, and it's not a problem if we load it from RSpec.

pirj:
@eregon Amazing, thank you!
@ioquatix Would `#owned?` fix your failing case?

ioquatix:
I can try it out this week and report back

lib/rspec/support/reentrant_mutex.rb
JonRowe: Yes, we don't require things that developers need to require because it causes weird failures for developers. (e.g. RSpec will pass because we require it but it won't work in dev/production).

lib/rspec/support/reentrant_mutex.rb
JonRowe: Would this not break in a more typical threaded solution? E.g. JRuby

eregon: >  I hope they don't do it dynamically when there are existing fibers and reentrant mutexes in action.

I don't think it's advisable to take that risk/assumption, and anyway AFAIK there is no need.

eregon:
I made a PR using `Mutex#owned?`: https://github.com/rspec/rspec-support/pull/503

ioquatix:
I defer to @eregon thanks for your help here.

PR #504, opened 2021-04-28
Use Mutex#owned? to correctly check if the Mutex is owned by the current Thread or Fiber

### LINKED ISSUE #501: ThreadError: Attempt to unlock a mutex which is locked by another thread/fiber

pirj:
In Ruby >= 3, Mutex are held per Fiber, not per Thread.

Fixes #501

This is a backport of #503 for RSpec 3

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,8 @@
+### Development
+
+Bug Fixes:
+* Fix reentrant mutex for Ruby 3.0. (Benoit Daloze, #504)
+
 ### 3.10.2 / 2021-01-28
 [Full Changelog](http://github.com/rspec/rspec-support/compare/v3.10.1...v3.10.2)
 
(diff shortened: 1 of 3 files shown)


pirj:
Green on the first try!

JonRowe (approved):
I tweaked the changelog for 4.0 but I've picked that across here.

pirj:
Thanks, Jon!

ioquatix:
Has this been released yet?

ioquatix:
Oh, was this released 3 days ago? I still seem to experience some hangs in some cases, so maybe I need to investigate.

PR #505, opened 2021-06-11
Support end-less methods in Source::Token

yujinakayama:
This fixes https://github.com/rspec/rspec-core/issues/2892.


pirj (approved):
Works perfectly. Thank you!

PR #517, opened 2021-09-05
Move dev dependencies to Gemfile

pirj:
Sibling PRs:
 - https://github.com/rspec/rspec-expectations/pull/1320
 - https://github.com/rspec/rspec-mocks/pull/1439
 -  https://github.com/rspec/rspec-core/pull/2877

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,4 +24,7 @@ if RUBY_VERSION >= '2.4' && RUBY_ENGINE == 'ruby'
   gem "rubocop", "~> 0.52.1"
 end
 
+gem "rake", ">= 12.3.3"
+gem "thread_order", "~> 1.1.0"
+
 eval_gemfile 'Gemfile-custom' if File.exist?('Gemfile-custom')
--- a/rspec-support.gemspec
+++ b/rspec-support.gemspec
@@ -33,7 +33,4 @@ Gem::Specification.new do |spec|
   end
 
   spec.required_ruby_version = '>= 2.3.0'
-
-  spec.add_development_dependency "rake", ">= 12.3.3"
-  spec.add_development_dependency "thread_order", "~> 1.1.0"
 end


JonRowe:
Do you have a citation saying this is now considered best practise? As I don't want to do this without some community backing of "this is a better idea ™"; Gems normally have an empty bundle file (other than `gemspec`). 

The only reason we do is our ridiculous whackamole for older Rubies. I'd much rather remove all the gem overrides and go back to relying on gemspecs. (Same goes for linked PRs)

pirj:
> Do you have a citation saying this is now considered best practise?

Not really. The only remotely related confirmation for this is https://bundler.io/v2.2/man/gemfile.5.html#GEMSPEC

> The .gemspec file is ... where you specify the dependencies your gem needs **to run**.

Even though it contradicts the very existence of `add_development_dependency`.

`.gemspec` is controller by `rubygems`, while `Gemfile` by `bundler`. I can't figure out a local development workflow without `bundler` these days.

If we ever happen to need to test against different versions of soft-dependencies (e.g. `mocha`), we could use `appraisal`-controller `Gemfile`s, and specify those versions there. And keep the `gemspec` clean.

> much rather remove all the gem overrides

Absolutely agree. We'd have to abandon JRuby tests, make `rubocop` an optional step for certain Ruby versions, and skip `minitest`. Sounds like a significant change. I'm sympathetic to JRuby, but it's probably a good moment for them to start running stable builds of RSpec in scope of their CI, not the other way around.

JonRowe:
As rubygems / bundler come closer to merging, maybe its worth asking the core team what they would do? We can then follow them and hey thats an opportunity for a blog post to talk about it :joy:

pirj:
Smells free virtual karma! 😆 

I'll close this, since due to curious circumstances, there's no clarity what approach is correct. For sibling PRs removed this change, too.

Only kept Cucumber update.
There is an update to script files, too, if you approve, I'll send script update PRs using `rspec-dev`, and will rebase cucumber update PRs.

pirj:
I'd love to revive this PR eventually, in the light of new circumstances, with [the introduction](https://github.com/rubocop/rubocop/pull/11469) of [`Gemspec/DevelopmentDependencies` cop](https://docs.rubocop.org/rubocop/1.51/cops_gemspec.html#enforcedstyle-gemfile-default).

JonRowe:
Citation still required, Rubocop has a ton of rules which aren't "standard"...

JonRowe:
I read that (took a while, very wordy) but it doesn't seem like a clear conclusion, I think until its marked as deprecated we'll stick with specifying them in the gemspec.

PR #521, opened 2021-12-05
Replace deadlink inside encoded_string_spec.rb

tiendo1011:


JonRowe:
Thanks for pointing this out, I've amended the link back to ruby/spec as it's intended to be a link to a specific commit within the spec not ruby itself.

PR #523, opened 2021-12-28
Updates from rspec-dev (2021-12-28)

JonRowe:


pirj:
I think this is fine, as it was only failing for RuboCop. Let's turn it off for Ruby 3.1 for now.

JonRowe:
I've closed it only because the tools will create a new branch :joy:

PR #525, opened 2022-01-09
Update to a version of Rubocop that is compatible with Ruby 2.4-3.1

petergoldstein:
This is an update to Rubocop that maintains compatibility with Ruby 2.4-3.1 using a common .rubocop_rspec_base.yml file across `rspec-core`, `rspec-expectations`, `rspec-mocks`, and `rspec-support`

Changes to this common base configuration file are minimal, limited to:

1. Adding missing Departments for Cops
2. Splitting/renaming Cop configurations where directed by Rubocop for pre-existing configurations.
3. Removing a Performance cop configuration as that is now packaged in a separate library.

Each repo now has a .rubocop_todo.yml file which can be worked through at the maintainers leisure.

Some updates to the existing .rubocop.yml files were made, mostly in removing duplication and updating some variables so Rubocop would run green.

I also added some respond_to? guards for two undef statements.


.rubocop.yml
```diff
@@ -1,12 +1,18 @@
...
+
+
+# Over time we'd like to get this down, but this is what we're at now.
 Metrics/AbcSize:
   Max: 28
 
+# Over time we'd like to get this down, but this is what we're at now.
 Metrics/BlockLength:
-  Max: 86
+  Max: 834
```
JonRowe, line 13: This is a very large increase for support, is this file just copied and pasted?

.rubocop.yml
```diff
@@ -1,12 +1,18 @@
 inherit_from:
   - .rubocop_rspec_base.yml
+  - .rubocop_todo.yml
 
+
+
```
JonRowe, line 6: No need to add extra whitespace.

.rubocop_rspec_base.yml
```diff
@@ -1,79 +1,76 @@
-# This file was generated on 2021-07-15T10:45:53+01:00 from the rspec-dev repo.
+# This file was generated on 2021-07-15T10:45:52+01:00 from the rspec-dev repo.
```
JonRowe, line 1: We generally update this file using rspec-dev that can make multiple changes at once, I'd prefer to do that and then add the extra "version bump" commit to those prs by hand.

.rubocop.yml
petergoldstein: No, it wasn't.  It was necessary to accomodate the blocks in `spec`.  Excluding those allows us to go back to the original value.  I've update the PR.

PR #526, opened 2022-01-09
Guard undefs and use an object that implements :=~ in all Ruby versions for specs

petergoldstein:
Starting in Ruby 3.2 arrays no longer implement `=~`. So add a respond_to? in the proxy definition for hygiene and use a different object (a string) that does implement that method.


pirj:
Thank you!

PR #533, opened 2022-02-09
Support TruffleRuby

gogainda:


pirj:
Wouldn't it be more convenient for TruffleRuby to include RSpec into their build?

pirj:
One reason for this is that we're really capped by the GHA tier, and with the number of Ruby versions we support getting a single build to run takes several minutes.
For `rspec-rails`, it's even worse - it can be up to half an hour. If other builds in `rspec` organization are running, they are queuing up.

To ba fair, we're seizing support for older Ruby versions in RSpec 4, but there are still [a plenty of versions to support](https://github.com/rspec/rspec-support/runs/4773349617?check_suite_focus=true), with Ruby 3.1 and `ruby-head` missing from that list.

We were thinking of removing JRuby builds and relying on JRuby to test Ruby compatibility on their side, but don't want to do this abruptly.
If there's an incompatibility with TruffleRuby or JRuby, it's a burden for us to fix, and we hope for the cooperation of their core teams to help with the fixes.

Do you think it would be possible to add a job that runs RSpec tests along TruffleRuby builds instead of adding a TruffleRuby job that runs along with RSpec builds?

.github/workflows/ci.yml
```diff
@@ -45,6 +45,7 @@ jobs:
           - 2.3
           - 2.2
           - 2.1.9
+          - truffleruby-head
```
JonRowe, line 48: We're not adding truffleruby to our CI matrix, please remove this.

lib/rspec/support/caller_filter.rb
```diff
@@ -19,13 +19,19 @@ class CallerFilter
 
     LIB_REGEX = %r{/lib/rspec/(#{(RSPEC_LIBS + ADDITIONAL_TOP_LEVEL_FILES).join('|')})(\.rb|/)}
 
+    LANGUAGE_SPECIFIC_REGEX = []
+    LANGUAGE_SPECIFIC_REGEX <<  "<internal:" if defined?(RUBY_ENGINE) && RUBY_ENGINE == 'truffleruby'
```
JonRowe, line 23: If we need to add specific things for truffle ruby we'd prefer to add a truffle ruby flag in our ruby module and use that rather than repeat checks everywhere.

JonRowe (changes requested):
Truffle ruby is not officially supported by RSpec, we're open to easy changes to support their efforts, see the feedback on this PR.

lib/rspec/support/caller_filter.rb
eregon: It would be safer to anchor the pattern though, so "^<internal:" instead of "<internal:"

spec/rspec/support/spec/stderr_splitter_spec.rb
```diff
@@ -86,7 +86,9 @@
     splitter.verify_no_warnings!
   end
 
-  unless defined?(RUBY_ENGINE) && RUBY_ENGINE == 'rbx'
+  unless defined?(RUBY_ENGINE) && (RUBY_ENGINE == 'rbx' || RUBY_ENGINE == 'truffleruby')
```
JonRowe, line 89: Please use this:
https://github.com/rspec/rspec-support/blob/main/lib/rspec/support/ruby_features.rb#L51

JonRowe (changes requested):

spec/rspec/support/spec/stderr_splitter_spec.rb
```diff
@@ -86,7 +86,9 @@
     splitter.verify_no_warnings!
   end
 
-  unless defined?(RUBY_ENGINE) && RUBY_ENGINE == 'rbx'
+  unless Ruby.rbx? || Ruby.truffleruby?
```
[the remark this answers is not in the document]
gogainda: yep, it's not needed here, will update

spec/rspec/support/spec/stderr_splitter_spec.rb
```diff
@@ -86,7 +86,9 @@
     splitter.verify_no_warnings!
   end
 
-  unless defined?(RUBY_ENGINE) && RUBY_ENGINE == 'rbx'
+  unless Ruby.rbx?
+    # TruffleRuby doesn't support warnings for now
+    # https://github.com/oracle/truffleruby/issues/2595
```
pirj, line 91: Leftover

spec/rspec/support/spec/stderr_splitter_spec.rb
```diff
@@ -86,7 +86,7 @@
     splitter.verify_no_warnings!
   end
 
-  unless defined?(RUBY_ENGINE) && RUBY_ENGINE == 'rbx'
+  unless Ruby.rbx?
```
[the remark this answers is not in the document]
pirj: Indeed, I didn't mean "please remove this". If you need this on TruffleRuby CI to make sure `rspec-*` jobs pass - please feel free to keep this.

pirj, line 89: I guess it's `RSpec::Support::Ruby.rbx?`.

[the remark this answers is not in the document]
eregon: Indeed, it's not because truffleruby is not currently tested in rspec-support's CI that it's not useful. It might be tested in some other CI and locally.

JonRowe (approved):
Assuming this passes the build, I'll squash merge this today.

eregon:
Thank you for merging this.
I opened an issue to figure what would be needed to officially support TruffleRuby in RSpec: https://github.com/rspec/rspec-metagem/issues/68

PR #536, opened 2022-02-22
Add test for failing case on Ruby 3

dorner:
For inspection for issue rspec/rspec#104.


JonRowe:
Closing during the monorepo migration as we're going to be looking at better keyword argument support in RSpec 4 and this doesn't migrate cleanly due to previous improvements.

Issue #522, opened 2021-11-03
Combination of option hash and keyword arguments causes crashes


dorner:
### Subject of the issue

When verified doubles are turned on, a particular combination of method arguments causes RSpec to crash - it gets confused with the method signature.

### Your environment
* Ruby version: 3.0
* rspec-mocks version: 3.10.2

### Steps to reproduce

Run the following file:

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'rspec'
end

require "rspec/autorun"

RSpec.configure do |config|
  config.mock_with :rspec do |mocks|
    mocks.verify_partial_doubles = true
  end
end

module MyModule
  def self.my_func(arg1, arg2, options={}, kw: false)
  end
end

RSpec.describe MyModule do
  it 'should pass' do
    allow(MyModule).to receive(:my_func).and_return('hi mom')
    MyModule.my_func(1, 2, { foo: 'bar', baz: 'eggs' })
  end
end
```

### Expected behavior
The spec should pass.

### Actual behavior
Fails with the following error:

```
     Failure/Error: MyModule.my_func(1, 2, { foo: 'bar', baz: 'eggs' })
     
     ArgumentError:
       Invalid keyword arguments provided: foo, baz
```

Note that if you replace the call by explicitly calling the keyword argument and not relying on the default, the spec passes: `MyModule.my_func(1, 2, { foo: 'bar', baz: 'eggs' }, kw: false)`

pirj:
Thanks for reporting, and apologies that it took long to process.

Apparently, there are two issues. One is in `MethodSignatureVerifier#has_kw_args_in?`, and it seems to be an easy fix:
```diff
      def has_kw_args_in?(args)
        Hash === args.last &&
          could_contain_kw_args?(args) &&
-         (args.last.empty? || args.last.keys.any? { |x| x.is_a?(Symbol) })
+         (args.last.empty? || args.last.keys.any? { |x| x.is_a?(Symbol) }) &&
+         (args.last.keys & @allowed_kw_args).any?
      end
```

Except that it's not correct.

When the last argument is passed in curly braces, it should **not** be treated as kwargs.
```ruby
> def my_func(x, y, options={}, kw: false)
>   puts options, kw
> end
> my_func(1,2, foo: 1, kw: 1) # ArgumentError: unknown keyword: :foo
> my_func(1,2, {foo: 1, kw: 1})
{:foo=>1}
false
```

[...]

PR #537, opened 2022-03-18
Distinct last positional argument hashes from keywords on Ruby 3

JonRowe:
This is a solution attempt for rspec/rspec-support#536 

On Ruby 3 we can (with effort) distinguish between a last positional hash and a keyword hash, because the mixed behaviour we can get on other Rubies shouldn't happen.

### DIFF
--- a/lib/rspec/support/method_signature_verifier.rb
+++ b/lib/rspec/support/method_signature_verifier.rb
@@ -85,12 +85,58 @@ module RSpec
             (args.last.empty? || args.last.keys.any? { |x| x.is_a?(Symbol) })
         end
 
-        # Without considering what the last arg is, could it
-        # contain keyword arguments?
         def could_contain_kw_args?(args)
+          # If the arg count is less than the required positional args there are no keyword args
           return false if args.count <= min_non_kw_args
 
-          @allows_any_kw_args || @allowed_kw_args.any?
+          # If there is no keyword splat or allowed keywords there are no keyword args
+          return false unless @allows_any_kw_args || @allowed_kw_args.any?
+
+          # If the args captured are more than the positional arguments then there are keywords
+          return true if args.count > max_non_kw_args
+
+          # Otherwise we need to check wether the last argument is a positional non default hash,
+          # e.g. that this:
+          #
+          # def capture(_, last = {}, keyword: 2)
+          #
+          # Is valid with both:
+          #
+          # [1, {:some => :hash}]
+          # [1, {:keyword => :value}]
+          #
+          # but for different reasons.
+
+          # If the keyword splat is allowed, and there are no specified
+          # keywords args then we will either have returned true on L100
+          # or this is invalid.
+          return false if @allowed_kw_args.empty? && @allows_any_kw_args
+
+          if Hash === args.last
+            last_arg_matches_any_keywords = !(args.last.keys & @allowed_kw_args).empty?
+            has_extra_keys = !(args.last.keys - @allowed_kw_args).empty?
+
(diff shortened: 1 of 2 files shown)


lib/rspec/support/method_signature_verifier.rb
```diff
@@ -85,12 +85,58 @@ def has_kw_args_in?(args)
             (args.last.empty? || args.last.keys.any? { |x| x.is_a?(Symbol) })
         end
 
-        # Without considering what the last arg is, could it
```
pirj, line 88: I have a faint feeling that the incorrect behaviour is in the `has_kw_args_in?` above, not related to the `could_contain_kw_args?` call.

With the test case that reproduces the problem from #522,
```ruby
            it 'treats the final positional argument as a hash' do
              expect(valid?(1, { foo: 'bar', baz: 'eggs'})).to eq(true)
            end
```

The problem is a call to `arity_kw` defined as:
```ruby
def arity_kw(x, y = {}, z:2); end
```
 - **may** contain kwargs (so arguably `could_contain_kw_args?` returning `true` is correct)
 - has a variable number of positional arguments

The last argument can be matched to either `y` or considered a kwargs depending on what this last argument is.

With `{ foo: 'bar', baz: 'eggs'}` specifically, we can really detect it's a kwarg by checking if:
1. it is a Hash
2. all of its keys are Symbols
3. **all** of its keys of the arg are in the method signature as kwargs

With `foo` and `bar` as the last arg keys and `z` as the only kwarg in the method signature, we can certainly tell that the last argument is **not** a kwarg.

lib/rspec/support/method_signature_verifier.rb
```diff
@@ -85,12 +85,58 @@ def has_kw_args_in?(args)
...
+            last_arg_matches_any_keywords = !(args.last.keys & @allowed_kw_args).empty?
+            has_extra_keys = !(args.last.keys - @allowed_kw_args).empty?
+
+            # If we allow the keyword splat, and the last hash contains
+            # a specified keyword, then this hash contains keywords
+            return true if @allows_any_kw_args && last_arg_matches_any_keywords
+
+            # If we don't allow the keyword splat, and the last hash contains extra keys
+            # then it is a positional hash.
+            return false if !@allows_any_kw_args && has_extra_keys
```
pirj, line 125: I guess there's a problem between 2.7 and 3.0+ with (`spec/rspec/support/method_signature_verifier_spec.rb:372`)
```ruby
> def arity_kw(x, y = {}, z:2); end

> arity_kw(nil, {a: 1}, {z: 2})
```
it works fine on 2.7 but fails on 3.0+:
```
ArgumentError: wrong number of arguments (given 3, expected 1..2)
```

To me, it seems that this expectation is not justified on Ruby 2.7:
```
expect(valid?(nil, :a => 1)).to eq(false)
```
as `y` is initialized with `{:a => 1}`.

I didn't look at other few failing examples yet.

lib/rspec/support/method_signature_verifier.rb
pirj: Sorry for the lengthy comment, I've tried to refresh my memory by writing things down. Glancing over, I see that your implementation is addressing this exact problem in `has_extra_keys`.

JonRowe: Yep this PR is entirely about addressing ambiguity where traditional style hashes can be keywords when there is this duality.

JonRowe:
Migrated to the monorepo as rspec/rspec#120

PR #541, opened 2022-04-13
Disable ripper in TruffleRuby due to slow performance

bjfish:
From comment:
https://github.com/rspec/rspec-metagem/issues/68#issuecomment-1098037326

> I think we should disable the usage of Ripper by RSpec for TruffleRuby, in practice it seems currently very slow (seconds of pause just before showing errors & backtraces) and so it hurts the user experience a lot more than help.

cc @eregon


pirj:
Thanks

PR #549, opened 2022-09-13
Synchronise changes from main / 3-11-maintenance to 4-0-dev

JonRowe:
This is mostly rubocop and typo fixes which have been sorted out on the other two branches and brings 4-0 closer to parity with main.


Gemfile
```diff
@@ -26,7 +26,7 @@ end
 
 # No need to run rubocop on earlier versions
 if RUBY_VERSION >= '2.4' && RUBY_ENGINE == 'ruby'
-  gem "rubocop", "~> 0.52.1"
+  gem 'rubocop', "~> 1.0", "< 1.12"
```
pirj, line 29: I think there was a bug in RuboCop preventing it from correctly parsing older Ruby code when running on newer Ruby versions or the other way around. I'd just remove `< 1.12` and see what happens.

pirj:
Green!

pirj (approved):
Thank you!

PR #553, opened 2022-11-18
Fix `ReentrantMutex` locking inside `Fiber` spec for ruby-3.2

ojab:
It raises `ThreadError` since https://github.com/ruby/ruby/pull/6680

### DIFF
--- a/spec/rspec/support/reentrant_mutex_spec.rb
+++ b/spec/rspec/support/reentrant_mutex_spec.rb
@@ -28,7 +28,18 @@ RSpec.describe RSpec::Support::ReentrantMutex do
     order.join_all
   end
 
-  if RUBY_VERSION >= '3.0'
+  if RUBY_VERSION >= '3.2'
+    it 'raises an error when trying to lock from another Fiber' do
+      mutex.synchronize do
+        Fiber.new do
+          expect {
+            mutex.send(:enter)
+            raise 'should not reach here: mutex is already locked on different Fiber'
+          }.to raise_error(ThreadError, 'deadlock; lock already owned by another fiber belonging to the same thread')
+        end.resume
+      end
+    end
+  elsif RUBY_VERSION >= '3.0'
     it 'waits when trying to lock from another Fiber' do
       mutex.synchronize do
         ready = false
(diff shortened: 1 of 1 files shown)


ojab:
>First-time contributors need a maintainer to approve running workflows. 

Sending to review, specs are green locally with ruby-3.2-preview3

ojab:
Also, `should reach here` is copypasted from existing spec, shouldn't it be `should not reach here`?

pirj:
The old spec [didn't fail](https://github.com/rspec/rspec-support/actions/runs/3331958379/jobs/5512429885) with `ruby 3.2.0dev (2022-10-26T15:29:12Z master fa0adbad92)`, and it [doesn't fail now](https://github.com/rspec/rspec-support/actions/runs/3502653809/jobs/5867184026) with `ruby 3.2.0dev (2022-11-17T18:58:56Z master 0446d961a0)`.
The [breaking change in Ruby](https://github.com/ruby/ruby/pull/6680) was made between those runs.
We [skip sub-builds for `ruby-head`](https://github.com/rspec/rspec-mocks/actions/runs/3498069669/jobs/5863512421), so it's not possible to confirm the failure from `rspec-mocks` build.

I intend to:
- stash the changes temporarily
- reproduce the failure
- get the changes back

pirj:
A [failing 3.2 build](https://github.com/rspec/rspec-support/actions/runs/3502699953/jobs/5867268699) 👍

pirj (approved):
Thank you!

JonRowe:
Realistically this should be a pending spec on Ruby 3.2 as its something we should look to fix... (Its not expected to error, it just does)...

eregon (approved):
Looks good to me, this just adapts the test to the new exception/behavior in CRuby

JonRowe:
I've been working on reworking the spec to run, I don't think its a change in behaviour it's just the way of testing it that changed.

eregon:
It's a behavior change in Ruby, the deadlock is detected and so it doesn't hang.
I think it's difficult to design a test that works with both and is still readable, so I think this is a good solution.

An alternative would be to use a Fiber scheduler, then it would behave as before, since with a scheduler it makes sense to wait. A Fiber scheduler is probably the reason we needed to fix RSpec ReentrantMutex. Would need to make a dummy Fiber scheduler for tests, it's probably fairly short if only used for this example.

eregon:
https://github.com/rspec/rspec-support/issues/501 is the original issue

JonRowe:
I'm closing this for now, we currently skip this on Rubies were its broken, I'd like to fix this spec rather than a no op expect an error (probably needing a fiber scheduler) especially as on my machine its definielty a broken spec (the raise leaks rather than a deadlock)

PR #559, opened 2023-01-06
chore: Add Ruby 3.2 to the CI matrix

ydah:
Follow up: rspec/rspec-rails#2648


pirj:
Thank you. There's a build failure related to mutex behaviour change that makes all 3.2 builds red.
Usually, we change build and some doc files in a batch in a centralized way from https://github.com/rspec/rspec-dev/ to keep them in sync.

Issue #554, opened 2022-11-23
String keys should be allowed in **kwargs


johnnymo87:
### Subject of the issue
Ruby 2.7 introduce the feature of [last-argument non-specified keyword argument `**kwargs`](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/#other-minor-changes-non-symbol-keys) in method signatures. In this particular kind of keyword argument, String keys **_are_** allowed. But in this library's `MethodSignature` logic, we're counting on keyword arguments of any kind always being Symbol keys ([1](https://github.com/rspec/rspec-support/blob/49b89c6af0408f543c0be3772553853ec3fc9efd/lib/rspec/support/method_signature_verifier.rb#L80-L86), [2](https://github.com/rspec/rspec-support/blob/49b89c6af0408f543c0be3772553853ec3fc9efd/lib/rspec/support/method_signature_verifier.rb#L363-L378)). So in the case of `**kwargs` with String keys, we misinterpret them and treat them as intended for a positional argument.

This issue is similar to https://github.com/rspec/rspec-support/issues/522, but I believe it's not addressed by the WIP https://github.com/rspec/rspec-support/pull/537, so I figured I'd open a separate issue to highlight it.

### Your environment
* Ruby version: 3.1
* rspec-support version: 3.11

### Steps to reproduce
Run the following file:
```rb
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'rspec'
end

require 'rspec/autorun'

RSpec.configure do |config|
  config.mock_with :rspec do |mocks|
    mocks.verify_partial_doubles = true
  end
end

module MyModule
  def self.my_func(arg1, **kwargs); end
end

RSpec.describe MyModule do
  it 'works with kwargs with String keys' do
    expect(MyModule).to receive(:my_func).with(1, 'foo' => 'bar')

    MyModule.my_func(1, 'foo' => 'bar')
  end
end
```

### Expected behavior
The spec should pass.

### Actual behavior
Fails with the following error:
```
     Failure/Error: expect(MyModule).to receive(:my_func).with(1, 'foo' => 'bar')
       Wrong number of arguments. Expected 1, got 2.
```
Note that this passes if I change `'foo' => 'bar'` to `:foo => 'bar'`.

JonRowe:
PR accepted to improve this

malcolmohare:
[...]

PR #591, opened 2024-01-15
Allow string keys for method kwarg splats

malcolmohare:
Attempting to fix: https://github.com/rspec/rspec-support/issues/554

The change expands the matching criteria for kwargs from `x.is_a?(Symbol)` to `x.is_a?(Symbol) || x.is_a?(String)`.  This now properly mirrors the behavior in the ruby docs.

https://docs.ruby-lang.org/en/3.0/syntax/calling_methods_rdoc.html#label-Hash+to+Keyword+Arguments+Conversion

Elsewhere tests were asserting that args with string keys were treated as optional arguments.  I have not tested if this is true for ruby prior to ruby 3, but this is NOT true for ruby 3, and I've adjusted those tests.

**Testing**
```
bundle exec rspec spec/rspec/support/method_signature_verifier_spec.rb
...
Finished in 0.03491 seconds (files took 0.09141 seconds to load)
268 examples, 0 failures
```
Running script/run_build fails with a permissions error during rspec, so I don't have a clean test pass of the whole repo.

### DIFF
--- a/lib/rspec/support/method_signature_verifier.rb
+++ b/lib/rspec/support/method_signature_verifier.rb
@@ -84,7 +84,7 @@ module RSpec
         def has_kw_args_in?(args)
           Hash === args.last &&
             could_contain_kw_args?(args) &&
-            (args.last.empty? || args.last.keys.any? { |x| x.is_a?(Symbol) })
+            (args.last.empty? || args.last.keys.any? { |x| x.is_a?(Symbol) || x.is_a?(String) })
         end
 
         # Without considering what the last arg is, could it
@@ -365,12 +365,12 @@ module RSpec
       def split_args(*args)
         kw_args = if @signature.has_kw_args_in?(args)
                     last = args.pop
-                    non_kw_args = last.reject { |k, _| k.is_a?(Symbol) }
+                    non_kw_args = last.reject { |k, _| k.is_a?(Symbol) || k.is_a?(String) }
                     if non_kw_args.empty?
                       last.keys
                     else
                       args << non_kw_args
-                      last.select { |k, _| k.is_a?(Symbol) }.keys
+                      last.select { |k, _| k.is_a?(Symbol) || k.is_a(String) }.keys
                     end
                   else
                     []
--- a/lib/rspec/support/method_signature_verifier.rb
+++ b/lib/rspec/support/method_signature_verifier.rb
@@ -364,14 +364,7 @@ module RSpec
 
       def split_args(*args)
         kw_args = if @signature.has_kw_args_in?(args)
-                    last = args.pop
-                    non_kw_args = last.reject { |k, _| k.is_a?(Symbol) || k.is_a?(String) }
-                    if non_kw_args.empty?
-                      last.keys
-                    else
-                      args << non_kw_args
-                      last.select { |k, _| k.is_a?(Symbol) || k.is_a(String) }.keys
-                    end
+                    args.pop.select { |k, _| k.is_a?(Symbol) || k.is_a?(String) }.keys
                   else
                     []
                   end
--- a/lib/rspec/support/method_signature_verifier.rb
+++ b/lib/rspec/support/method_signature_verifier.rb
@@ -84,7 +84,7 @@ module RSpec
         def has_kw_args_in?(args)
           Hash === args.last &&
             could_contain_kw_args?(args) &&
-            (args.last.empty? || args.last.keys.any? { |x| x.is_a?(Symbol) || x.is_a?(String) })
+            (args.last.empty? || args.last.keys.any? { |x| x.is_a?(Symbol) || RubyFeatures.kw_arg_separation? })
         end
 
         # Without considering what the last arg is, could it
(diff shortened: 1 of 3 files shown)


lib/rspec/support/method_signature_verifier.rb
```diff
@@ -365,12 +365,12 @@ def unlimited_args?
...
         kw_args = if @signature.has_kw_args_in?(args)
                     last = args.pop
-                    non_kw_args = last.reject { |k, _| k.is_a?(Symbol) }
+                    non_kw_args = last.reject { |k, _| k.is_a?(Symbol) || k.is_a?(String) }
                     if non_kw_args.empty?
                       last.keys
                     else
                       args << non_kw_args
-                      last.select { |k, _| k.is_a?(Symbol) }.keys
+                      last.select { |k, _| k.is_a?(Symbol) || k.is_a(String) }.keys
```
pirj, line 373: Question mark is missing.
This makes me think this branch is not covered in tests. Can you please check this?

malcolmohare: I think the whole chunk of code is unnecessary for the current ruby version.  I went through the [PR](https://github.com/rspec/rspec-support/pull/366) that added it, and the parameter declarations that PR was written to support no longer function (at least in my irb console which is ruby 3).

Example:
```
def foo(x, y = {}, z: 1) puts "#{y}, #{z}"; end
foo 1, 'a' => 2, 'b' => 2
ArgumentError (unknown keywords: "a", "b")
```

So i'm not sure what the correct thing to do here is.  Do I update the code to respect ruby 3 functionality?  Or does there need to be some if statement on the runtime version in the code?

spec/rspec/support/method_signature_verifier_spec.rb
```diff
@@ -384,17 +384,21 @@ def arity_kw(x, y = {}, z:2); end
...
+            it 'treats symbols as keyword arguments' do
+              expect(valid?(nil, :z => 3)).to eq(true)
+              expect(valid?(nil, :b => 3)).to eq(false)
+              expect(valid?(nil, :b => 2, :z => 3)).to eq(false)
+            end
+
+            it 'treats string keys as invalid keyword arguments' do
+              expect(valid?(nil, 'a' => 1)).to eq(false)
+              expect(valid?(nil, 'a' => 1, :z => 3)).to eq(false)
             end
```
JonRowe, line 396: This needs to be changed to conditonally apply the new behaviour, the old spec was valid for Ruby 2.x

Typically we'd define whole different specs for different Rubies.

spec/rspec/support/method_signature_verifier_spec.rb
```diff
@@ -384,17 +384,21 @@ def arity_kw(x, y = {}, z:2); end
...
 
             it 'mentions the invalid keyword args in the error', :pending => RSpec::Support::Ruby.jruby? && !RSpec::Support::Ruby.jruby_9000? do
               expect(error_for(1, 2, :a => 0)).to eq("Invalid keyword arguments provided: a")
               expect(error_for(1, :a => 0)).to eq("Invalid keyword arguments provided: a")
-              expect(error_for(1, 'a' => 0, :b => 0)).to eq("Invalid keyword arguments provided: b")
+              expect(error_for(1, 'a' => 0, :b => 0)).to eq("Invalid keyword arguments provided: a, b")
```
JonRowe, line 401: As above

JonRowe (changes requested):
This needs to be changed for Ruby 2.x

spec/rspec/support/method_signature_verifier_spec.rb
malcolmohare: Sure, added this as a RubyFeatures method and applied the change and the specs based on that flag.

lib/rspec/support/method_signature_verifier.rb
pirj: We can conditionally define methods depending on the Ruby version/features.

Does the “any key type counts as an optional kwargs key” apply to all Ruby versions that suport optional kwargs?

If there’s a conditional branch that isn’t used as we’ve discovered, let’s double check why it was introduced initially. I’d remove if it’s dead code.

malcolmohare: This section was re-written so these comments are a bit outdated.

But in reference to what you are asking, yes the logic needs to stay for ruby < 3, and the proposed change is correct for ruby >= 3.  The latest version of this PR introduces a RubyFeatures method to allow for this, and a link to the ruby changelog explaining the change.

lib/rspec/support/method_signature_verifier.rb
```diff
@@ -363,7 +363,7 @@ def unlimited_args?
       end
 
       def split_args(*args)
```
malcolmohare, line 365: There's an alternative version of this function:

```
def split_args(*args)
        return [args.length, kw_args] unless @signature.has_kw_args_in?(args)
        kw_args = if RubyFeatures.kw_arg_separation?
                            args.pop.keys
                          else
                            last = args.pop
                            non_kw_args = last.reject { |k, _| k.is_a?(Symbol) }
                            if non_kw_args.empty?
                              last.keys
                            else
                              args << non_kw_args
                              last.select { |k, _| k.is_a?(Symbol) }.keys
                           end
                         end
        [args.length, kw_args]
      end
    end
```

JonRowe:
Thank you for this 👏

JonRowe:
Released in 3.12.2 thank you!

PR #594, opened 2024-02-25
[bugfix] Properly detect kwargs hashes vs optional positional args

malcolmohare:
Attempting to fix https://github.com/rspec/rspec-expectations/issues/1451

Refactor `has_kw_args_in?`:

- try to use positional parameter information to determine whether a hash is kw_args
- try to leverage information passed by the ruby2_keywords annotation
- fallback to inspection of keyword args to determine if they look valid or not

Added backwards compatible ruby2_keywords wrapping to the necessary methods (valid? and error_for) in the spec files like so:
`ruby2_keywords(:valid?) if respond_to?(:ruby2_keywords, true)`

Also needed to modify how args are passed to the MSV split_args so that the special kwargs markings are not lost.

### DIFF
--- a/lib/rspec/support/method_signature_verifier.rb
+++ b/lib/rspec/support/method_signature_verifier.rb
@@ -80,7 +80,7 @@ module RSpec
         end
 
         # Considering the arg types, are there kw_args?
-        def has_kw_args_in?(args)
+        def has_kw_args_in?(args) # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
           if RubyFeatures.kw_arg_separation?
             # If the last arg is a hash, depending on the signature it could be kw_args or a positional parameter.
             return false unless Hash === args.last && could_contain_kw_args?(args)
--- a/lib/rspec/support/method_signature_verifier.rb
+++ b/lib/rspec/support/method_signature_verifier.rb
@@ -79,12 +79,35 @@ module RSpec
           given_kw_args - @allowed_kw_args
         end
 
-        # If the last argument is Hash, Ruby will treat only symbol keys as keyword arguments
-        # the rest will be grouped in another Hash and passed as positional argument.
+        # Considering the arg types, are there kw_args?
         def has_kw_args_in?(args)
-          Hash === args.last &&
-            could_contain_kw_args?(args) &&
-            (RubyFeatures.kw_arg_separation? || args.last.empty? || args.last.keys.any? { |x| x.is_a?(Symbol) })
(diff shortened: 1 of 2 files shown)


lib/rspec/support/method_signature_verifier.rb
```diff
@@ -79,12 +79,31 @@ def invalid_kw_args_from(given_kw_args)
...
-        # the rest will be grouped in another Hash and passed as positional argument.
+        # Considering the arg types, are there kw_args?
         def has_kw_args_in?(args)
-          Hash === args.last &&
-            could_contain_kw_args?(args) &&
-            (RubyFeatures.kw_arg_separation? || args.last.empty? || args.last.keys.any? { |x| x.is_a?(Symbol) })
+          if RubyFeatures.kw_arg_separation?
+            # If the last arg is a hash, depending on the signature it could be kw_args or a positional parameter.
+            return false unless Hash === args.last && could_contain_kw_args?(args)
+
```
pirj, line 87:
```suggestion

```
Fix for
```
lib/rspec/support/method_signature_verifier.rb has spaces on the EOL on lines 87
```

pirj:
Fails on Ruby 3+

malcolmohare:
Fully backwards compatible, just need to fix 3.2 and up.  The behavior of ruby2_keywords changed between 3.1 and 3.2: https://rubyreferences.github.io/rubychanges/3.2.html#keyword-argument-separation-leftovers

lib/rspec/support/method_signature_verifier.rb
```diff
@@ -282,7 +305,7 @@ class MethodSignatureVerifier
 
       def initialize(signature, args=[])
         @signature = signature
-        @non_kw_args, @kw_args = split_args(*args)
+        @non_kw_args, @kw_args = split_args(args.clone)
         @min_non_kw_args = @max_non_kw_args = @non_kw_args
         @arbitrary_kw_args = @unlimited_args = false
       end
```
malcolmohare, line 311: something had to change here.  Splatting the args meant the special marking for the kwargs hash gets lost.  This doesnt happen to functions with the ruby2_keywords... except it won't apply here because this function doesn't take the *args input.  So either the args array is cloned (which persists the special flagging) or the input to this function needs to swap to using *args.  I figured cloning was the less obtrusive change.

JonRowe:
Apologies I never got round to looking at this, I've migrated it over to the monorepo as rspec/rspec#121 I'm hoping to do a big refactoring of keyword argument support with RSpec 4 dropping older ruby support.

PR #596, opened 2024-04-25
Fixes diff when fuzzy finder `anything` is used in a Hash object (proof of concept)

KarlHeitmann:
Hi

I'd like to give you some context about this PR.

## TL; DR

See #551 , it is annoying to see the noise added by `anything` matcher on diffs and I wanted to fix that. I initially thought the solution was trivial, but I quickly found out I was wrong when I began reading the `Diff` class: there is a lot happening under the hood. I know there are some limitations to the Differ, and I understand there is a lot to change. This PR is a proposal, I don't know if my solution is too hacky, and I'd like to receive some feedback in order to solve this problem. 

Before:
```
@@ -1,4 +1,4 @@
-:an_key => anything,
+:an_key => "dummy",
 :fixed => "fixed",
-:nested => {:anything_key=>anything, :name=>"foo", :trigger=>"wrong"},
+:nested => {:anything_key=>"bcdd0399-1cfe-4de1-a481-ca6b17d41ed8", :name=>"foo", :trigger=>"trigger"},
```

After:
```
@@ -1,4 +1,4 @@
 :an_key => "dummy",
 :fixed => "fixed",
-:nested => {:anything_key=>"bcdd0399-1cfe-4de1-a481-ca6b17d41ed8", :name=>"foo", :trigger=>"wrong"},
+:nested => {:anything_key=>"bcdd0399-1cfe-4de1-a481-ca6b17d41ed8", :name=>"foo", :trigger=>"trigger"},
```

## Context

I am currently working on a RoR project that uses Jbuilder to render JSON data that I am delivering on the endpoints of my API.

I am testing my `.jbuilder` files with `type: view` RSpec tests, and I have a lot of `.json` fixture files I am loading and using on my test files as expected values.

## The problem this PR addresses.

My plan is to use these fixture files to kill two birds with one stone. On one side, I am comparing those fixture files with the JSON rendered by the `.jbuilder` files, and on the other side I am using those fixture files as mocks to share with the frontend developers, so they will have the data necessary to write their React application.

The plan is: if somebody in the backend changes a jbuilder, the tests will complain and I will remember to change the corresponding JSON fixture. This way, I will always have the JSON fixtures used by the frontend developers in sync with the backend rendered files. Frontend developers often need to work with data they don't know how to create on their local machines (because they do know React or Vue, but don't know the Rails framework)

In order to achieve the results of my plan, I ~ needed ~ to scrub some fields of my JSON fixture files using the `anything` fuzzy matcher. An example of fields I often need to scrub are the `id`, `created_at` and `updated_at` values.

I quickly discovered this annoying issue described on #551 : whenever I missed something that should make the test fail and render the diff in STDOUT, I found out the key-value pairs in my expected var with a `anything` matcher will add noise to the diff message, obfuscating which key actually differs from the expectation.

 ## My approach.
 
 My first idea was: _before_ the place in the source code that renders the diff string between `actual` and `expected` vars, I can add a stage where I'll search _every_ key-value pair that has an `anything` object as value in the `expected` variable, and _replace_ the `anything` value with the value found on the `actual` var I am testing.
 
 Hopefully the commits of this PR will show you what I mean. First commit consists in an implementation that will work with a Hash with no nested hashes, second commit will use recursivity to extend the solution to a hash object with nested hashes. 
 
## My reasons.

When I started writing the code to implement my idea, I've got a lot of questions: Can I just mutate the expected var? Shouldn't I get a notice message by the diff remembering that I've used the `anything` fuzzy matcher in my test? can I _just copy_ the actual value in the expected var? Am I lying when I am doing that?

My answer to these questions was: If I _decide_ to use the fuzzy matcher `anything` on a key-value pair in a hash object in my test, it implies I _do not care about seeing the `anything` in my diff_ . Because it is a wild card, because the `anything` fuzzy matcher should _morph_ into the real value on my actual variable.

I know this solution is hacky.... but it works! It gets rid of the noise generated by the `anything` fuzzy matcher. I tried to use the [super_diff](https://github.com/mcmire/super_diff)  gem, but it has the same problem.

I am sure the code I wrote can be improved and be more concise (as surely could be the text on this PR, lol), and I thought about extending the fix to cover `Array` objects. But I wanted to know your opinion about this issue before continuing with this. Can I use a recursive function this way? What impact on performance will have my lines of code in this extra stage I am adding? As Aristotle said: every scientific investigation should begin by gathering the opinion of the wise people (ie, the people that earlier on thought and worked on the problem I will begin to work right now).

I always liked the "diff" concept. And I am willing to continue collaborating on enhancing the diff tool. I've seen @mcmire wrote [this guide about RSpec](https://github.com/mcmire/super_diff/blob/main/docs/contributors/architecture/how-rspec-works.md#what-rspec-does), but I have not digged on the article yet. I am doing baby steps. Every feedback is appreciated.

Best.

### DIFF
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -17,13 +17,9 @@ module RSpec
 
         unless actual.nil? || expected.nil?
           if all_strings?(actual, expected)
-            if any_multiline_strings?(actual, expected)
-              diff = diff_as_string(coerce_to_string(actual), coerce_to_string(expected))
-            end
+            diff = diff_as_string(coerce_to_string(actual), coerce_to_string(expected)) if any_multiline_strings?(actual, expected)
           elsif hash_with_anything?(expected)
-            if no_procs_no_numbers.call(actual, expected)
-              diff = diff_as_object_with_anything(actual, expected)
-            end
+            diff = diff_as_object_with_anything(actual, expected) if no_procs_no_numbers.call(actual, expected)
           elsif no_procs_no_numbers.call(actual, expected)
             diff = diff_as_object(actual, expected)
           end
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -11,21 +11,9 @@ module RSpec
     # rubocop:disable Metrics/ClassLength
     class Differ
       def diff(actual, expected)
-        diff = ""
+        return "" if actual.nil? && expected.nil?
 
-        no_procs_no_numbers = lambda {|var1, var2| no_procs?(var1, var2) && no_numbers?(var1, var2)}
-
-        unless actual.nil? || expected.nil?
-          if all_strings?(actual, expected)
-            diff = diff_as_string(coerce_to_string(actual), coerce_to_string(expected)) if any_multiline_strings?(actual, expected)
-          elsif (keys = all_keys_from_hash(expected)) && keys.any?
-            diff = diff_as_object_with_anything(keys, actual, expected) if no_procs_no_numbers.call(actual, expected)
-          elsif no_procs_no_numbers.call(actual, expected)
-            diff = diff_as_object(actual, expected)
-          end
-        end
-
-        diff.to_s
+        run(actual, expected)
       end
 
       # rubocop:disable Metrics/MethodLength
@@ -88,6 +76,21 @@ module RSpec
       end
 
     private
+      def run(actual, expected)
+        diff = ""
+
+        no_procs_no_numbers = lambda {|var1, var2| no_procs?(var1, var2) && no_numbers?(var1, var2)}
+
+        if all_strings?(actual, expected)
+          diff = diff_as_string(coerce_to_string(actual), coerce_to_string(expected)) if any_multiline_strings?(actual, expected)
+        elsif (keys = all_keys_from_hash(expected)) && keys.any?
+          diff = diff_as_object_with_anything(keys, actual, expected) if no_procs_no_numbers.call(actual, expected)
+        elsif no_procs_no_numbers.call(actual, expected)
+          diff = diff_as_object(actual, expected)
+        end
+
+        diff.to_s
+      end
 
       def all_keys_from_hash(arg)
         return false unless Hash === arg
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -18,8 +18,8 @@ module RSpec
         unless actual.nil? || expected.nil?
           if all_strings?(actual, expected)
             diff = diff_as_string(coerce_to_string(actual), coerce_to_string(expected)) if any_multiline_strings?(actual, expected)
-          elsif hash_with_anything?(expected)
-            diff = diff_as_object_with_anything(actual, expected) if no_procs_no_numbers.call(actual, expected)
+          elsif (keys = all_keys_from_hash(expected)) && keys.any?
+            diff = diff_as_object_with_anything(keys, actual, expected) if no_procs_no_numbers.call(actual, expected)
           elsif no_procs_no_numbers.call(actual, expected)
             diff = diff_as_object(actual, expected)
           end
@@ -58,8 +58,8 @@ module RSpec
       end
       # rubocop:enable Metrics/MethodLength
 
-      def diff_as_object_with_anything(actual, expected)
-        @keys_with_anything.each do |keys|
+      def diff_as_object_with_anything(total_keys, actual, expected)
+        total_keys.each do |keys|
           pointer_expected = expected
           pointer_actual = actual
           final_key = keys.pop
@@ -89,11 +89,10 @@ module RSpec
 
     private
 
-      def hash_with_anything?(arg)
+      def all_keys_from_hash(arg)
         return false unless Hash === arg
 
-        @keys_with_anything = recursive_get_keys(arg)
-        @keys_with_anything.any?
+        recursive_get_keys(arg)
       end
 
       def recursive_get_keys(hash)
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -18,6 +18,10 @@ module RSpec
             if any_multiline_strings?(actual, expected)
               diff = diff_as_string(coerce_to_string(actual), coerce_to_string(expected))
             end
+          elsif Hash === expected && hash_with_anything?(expected)
+            if no_procs?(actual, expected) && no_numbers?(actual, expected)
+              diff = diff_as_object_with_anything(actual, expected)
+            end
           elsif no_procs?(actual, expected) && no_numbers?(actual, expected)
             diff = diff_as_object(actual, expected)
           end
@@ -56,6 +60,13 @@ module RSpec
       end
       # rubocop:enable Metrics/MethodLength
 
+      def diff_as_object_with_anything(actual, expected)
+        expected.select { |_, v| RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === v }.each_key do |k|
+          expected[k] = actual[k]
+        end
+        diff_as_object(actual, expected)
+      end
+
       def diff_as_object(actual, expected)
         actual_as_string = object_to_string(actual)
         expected_as_string = object_to_string(expected)
@@ -72,6 +83,10 @@ module RSpec
       end
 
     private
+
+      def hash_with_anything?(arg)
+        safely_flatten(arg).any? { |a| RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === a }
+      end
 
       def no_procs?(*args)
         safely_flatten(args).none? { |a| Proc === a }
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -97,9 +97,10 @@ module RSpec
       end
 
       def recursive_get_keys(hash)
-        klass = RSpec::Mocks::ArgumentMatchers::AnyArgMatcher
+        return [] unless defined?(RSpec::Mocks::ArgumentMatchers::AnyArgMatcher)
+
         hash.reduce([]) do |acc, pair|
-          if klass === pair[1]
+          if RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === pair[1]
             acc << [pair[0]]
           else
             if Hash === pair[1]
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -13,16 +13,18 @@ module RSpec
       def diff(actual, expected)
         diff = ""
 
+        no_procs_no_numbers = lambda {|var1, var2| no_procs?(var1, var2) && no_numbers?(var1, var2)}
+
         unless actual.nil? || expected.nil?
           if all_strings?(actual, expected)
             if any_multiline_strings?(actual, expected)
               diff = diff_as_string(coerce_to_string(actual), coerce_to_string(expected))
             end
           elsif hash_with_anything?(expected)
-            if no_procs?(actual, expected) && no_numbers?(actual, expected)
+            if no_procs_no_numbers.call(actual, expected)
               diff = diff_as_object_with_anything(actual, expected)
             end
-          elsif no_procs?(actual, expected) && no_numbers?(actual, expected)
+          elsif no_procs_no_numbers.call(actual, expected)
             diff = diff_as_object(actual, expected)
           end
         end
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -61,8 +61,15 @@ module RSpec
       # rubocop:enable Metrics/MethodLength
 
       def diff_as_object_with_anything(actual, expected)
-        expected.select { |_, v| RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === v }.each_key do |k|
-          expected[k] = actual[k]
+        @keys_with_anything.each do |keys|
+          pointer_expected = expected
+          pointer_actual = actual
+          final_key = keys.pop
+          keys.each do |k|
+            pointer_expected = pointer_expected[k]
+            pointer_actual = pointer_actual[k]
+          end
+          pointer_expected[final_key] = pointer_actual[final_key]
         end
         diff_as_object(actual, expected)
       end
@@ -85,7 +92,26 @@ module RSpec
     private
 
       def hash_with_anything?(arg)
-        safely_flatten(arg).any? { |a| RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === a }
+        @keys_with_anything = recursive_get_keys(arg)
+        @keys_with_anything.any?
+      end
+
+      def recursive_get_keys(hash)
+        klass = RSpec::Mocks::ArgumentMatchers::AnyArgMatcher
+        hash.reduce([]) do |acc, pair|
+          if klass === pair[1]
+            acc << [pair[0]]
+          else
+            if Hash === pair[1]
+              keys = recursive_get_keys(pair[1])
+              keys.each do |key|
+                key.prepend pair[0]
+                acc << key
+              end
+            end
+          end
+          acc
+        end
       end
 
       def no_procs?(*args)
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -105,7 +105,7 @@ module RSpec
             if Hash === pair[1]
               keys = recursive_get_keys(pair[1])
               keys.each do |key|
-                key.prepend pair[0]
+                key.unshift pair[0]
                 acc << key
               end
             end
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -18,7 +18,7 @@ module RSpec
             if any_multiline_strings?(actual, expected)
               diff = diff_as_string(coerce_to_string(actual), coerce_to_string(expected))
             end
-          elsif Hash === expected && hash_with_anything?(expected)
+          elsif hash_with_anything?(expected)
             if no_procs?(actual, expected) && no_numbers?(actual, expected)
               diff = diff_as_object_with_anything(actual, expected)
             end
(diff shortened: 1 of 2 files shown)


KarlHeitmann:
I've submitted 3 new commits to get rid of the CyclomaticComplexity, PerceivedComplexity & MethodLength cops. Each commit fixes one offense. The commits can be squashed into one commit, but since there are maaany ways to solve these offenses, these are only proposals.

KarlHeitmann:
Latest commit is to fix a problem when using ruby 2.4, 2.3, 2.2. It substitutes `prepend` method of `Array` class by `unshift`. Because prior ruby 2.4 it was not defined.

I can convert that last commit into a fixup to amend: [e4ece3df8e411ea4dab6eb08e02269ccf9c3fb14](https://github.com/rspec/rspec-support/pull/596/commits/e4ece3df8e411ea4dab6eb08e02269ccf9c3fb14)

KarlHeitmann:
Build is still failing before commit [f9a9e9d](https://github.com/rspec/rspec-support/pull/596/commits/f9a9e9d222115fa3e4150f728e5bb217302bef10). 

image

I don't know how to reproduce these build environment. I don't know what `ruby-head` stands for neither I understand the error there. And I don't know how to reproduce the 1.8.7 neither REE (what does this mean?) ruby versions in my computer to dig in and understand why it is failing.

However, I copied and pasted the ruby strings in an `irb` session and printed them to STDOUT using `puts`:

image

And found out on version 1.8.7 and REE the keys were not in _alphabetical_ order. I know `rspec-support` uses pretty print to make the diff of the objects, and my hypothesis is on ruby version 1.8.7, pretty print _maybe_ was not working correctly with some ruby hashes, and its output may be somewhat random.

The only solution I thought of was to _cheat_ and rearrange my tests so the keys are mimicking the output. Hopefully tests will pass now in version 1.8.7 and REE.

Any other ideas are welcome to make those ruby builds pass.

lib/rspec/support/differ.rb
```diff
@@ -73,6 +89,31 @@ def initialize(opts={})
 
     private
 
+      def hash_with_anything?(arg)
+        return false unless Hash === arg
+
+        @keys_with_anything = recursive_get_keys(arg)
```
pirj, line 95: Is it safe? Can we happen to reuse an i stance of a differ?

lib/rspec/support/differ.rb
```diff
@@ -73,6 +89,31 @@ def initialize(opts={})
...
+        @keys_with_anything.any?
+      end
+
+      def recursive_get_keys(hash)
+        klass = RSpec::Mocks::ArgumentMatchers::AnyArgMatcher
```
pirj, line 100: What if this is not defined, like someone opted out of mocking completely, or uses rr/mocha etc?

pirj:
Please accept my apologies for not reviewing this swiftly.

May I ask you to add the output of some spec how it looked before and after this change?

Don’t worry about ruby-head, it is very typical for it to fail. We mostly use it a bit ahead of time to fix issues with upcoming Ruby releases.

Thanks for the effort you’re putting into this.

lib/rspec/support/differ.rb
pirj: I think it’s ok to use ‘return [] unless defined?(RSpec::Mocks)’ here

lib/rspec/support/differ.rb
KarlHeitmann: That's a good question. On my opinion, I think nothing is ever safe :)

Let me see if I understand you. If we happen to reuse an instance of a differ the code will look like this snippet:

```suggestion
        @keys_with_anything = keys(arg)
      ...
      end
      
      def keys(arg) # NOTE: method originally named "recursive_get_keys"
        klass = RSpec::Mocks::ArgumentMatchers::AnyArgMatcher
        ...
          else
            if Hash === pair[1]
              keys = new.keys(pair[1]) # <- NOTE: here I will instance another Differ
              ...
        ...
      end
```

What I like about your idea, is by using a new instance of a Differ, we may benefit of mechanisms used by Differ class to rescue an error that may appear. We can leverage the power of a class in our favor, instead of recursively calling a single method.

The downside is currently the `hash_with_anything?` and `recursive_get_keys` methods are private. So I'll have to make them public. In addition, the `Differ` class may not be the appropriate class to get the keys of a Hash, maybe we should think about creating anothe file, eg: `Rspec::Support::Utils` that will have a method to retrieve keys recursively from a Hash, so we can maintain the single responsibility principle.

I googled a way to get all keys and nested keys of a Hash in Ruby and found this[ stack overflow thread](https://stackoverflow.com/questions/20235206/ruby-get-all-keys-in-a-hash-including-sub-keys). Most answers use recursive functions to solve the problem. There is no implementation in the ruby stdlib Hash class to get all keys (including nested keys). Maybe because the solution of this problem is not safe, so it is up to the programmer to do the implementation? I don't know. 

On my opinion, this may be unsafe. But since the _only_ condition to recursively call the `recursive_get_keys` method is [the value of the hash is a Hash](https://github.com/KarlHeitmann/rspec-support/blob/f9a9e9d222115fa3e4150f728e5bb217302bef10/lib/rspec/support/differ.rb#L105), then maybe this will narrow down the odds something will go wrong.

I think by reusing an instance of a differ is moving the recursion to another place. Is it better to move it to another place? I don't know. I will do what someone with more experience than me will advise me to do :)

If you think the change may be too risky, I can make the changes so we won't use recursion. After all, this project is used by 622K persons, I understand you need to be cautious.

lib/rspec/support/differ.rb
KarlHeitmann: Nice! I didn't thought someone may opt out mocking completely. I've fixed it on my latest commit!

KarlHeitmann:
Hey @pirj ! Thanks for reaching out! No worries, it is just my hobby try to fix simple things in github. I have on my bucket list to make a tiny contribution to an important project (like this one). But I understand people may be busier than me :rofl: 

I made [this gist with the output of these 3 specs BEFORE](https://gist.github.com/KarlHeitmann/e8fd6506eec82513656e2ea484301171) the changes here, and [this gist with the output AFTER](https://gist.github.com/KarlHeitmann/460639ccf4de0ec249f1d34b300ebe56) the changes. I hope this clarifies what I intend to do with my change.

My intention is whenever you want to diff a Hash with another Hash, if the _expected_ hash contains an [anything](https://github.com/rspec/rspec-mocks/blob/79877dd446e71b0b58d42ccf0b3616cae78fe05c/lib/rspec/mocks/argument_matchers.rb#L159) fuzzy matcher, that `anything` value will morph into the value of the actual variable, in order to reduce the noise generated by the diff if another key-value pair has a mismatch. As described here #551 

The three examples I wrote on my gists shows you one example of two hash comparisons with NO nested hashes, and the other two shows you what happens if there is a nested hash. Related to [your first comment](https://github.com/rspec/rspec-support/pull/596#discussion_r1593423984), if you think it is too unsafe to perform this recursively, I can tweak my PR so this will work only with plain hashes. The first commit of this PR implements this behavior ONLY on hash comparisons without nested hashes.

KarlHeitmann:
P.S.1: About support for ruby 1.8.7, I saw you've used `if String.method_defined?(:encoding) ... else ... end` block to define methods for ruby 1.8.7 and the other ruby versions dynamically. Maybe I can give it a try to fix the ruby build problem. 

P.S.2: I noticed something bad while using the debugger... after I mutate the hash on the expected `anything` value [here](https://github.com/KarlHeitmann/rspec-support/blob/f9a9e9d222115fa3e4150f728e5bb217302bef10/lib/rspec/support/differ.rb#L70), when PC returns to the `it` scope, the `expected` variable has mutated! As you can see on the screenshot below:
image
Line 611 has mutated the `expected[:an_key]` var from `anything` to `dummy`. I didn't notice that previously, I think this needs to be addressed.

lib/rspec/support/differ.rb
pirj: Can this be done with an accumulator passed to the recursive method instead of an instance variable?

pirj:
I recall something has changed in 1.9 or 2.0 related to the stability of keys order on hashes. This may cause the ordering issue.

Please correct me if I’m wrong, but this seems to only affect the order of pair in the output diff, and won’t cause any other behavioural difference between rspec runnin on 1.8.7 and later. If so, I suggest disregarding this problem, and marking the spec as broken on 1.8.7.

lib/rspec/support/differ.rb
KarlHeitmann: I am sorry, I don't know how that can be achieved.

KarlHeitmann:
Yes @pirj , it only affects the order of pair in the output diff. I'll mark the spec as broken on 1.8.7

lib/rspec/support/differ.rb
```diff
@@ -13,12 +13,14 @@ class Differ
...
+        no_procs_no_numbers = lambda {|var1, var2| no_procs?(var1, var2) && no_numbers?(var1, var2)}
+
         unless actual.nil? || expected.nil?
           if all_strings?(actual, expected)
-            if any_multiline_strings?(actual, expected)
-              diff = diff_as_string(coerce_to_string(actual), coerce_to_string(expected))
-            end
-          elsif no_procs?(actual, expected) && no_numbers?(actual, expected)
+            diff = diff_as_string(coerce_to_string(actual), coerce_to_string(expected)) if any_multiline_strings?(actual, expected)
+          elsif hash_with_anything?(expected)
```
pirj, line 21: To reset the discussion below.
I am concerned about side effects of shared state in instance variables.
I’d live to get rid of an instance variable `@keys_with_anything`.
Only those two methods need it. Can we cache it somehow here in a regular variable? Can `hash_with_anything?` become something that returns a value that we can subsequently pass to `diff_as_object_with_anything`?

KarlHeitmann:
Ok! I think I've solved the issues now. Please tell me if there is anything else I can do. I can also squash all the commits in one or two, but I think it is easier to review the code changes when each commit addresses a code suggestion change.

spec/rspec/support/differ_spec.rb
```diff
@@ -555,6 +556,75 @@ def inspect; "<BrokenObject>"; end
...
+              it "outputs only key value pair that triggered diff, anything_key should absorb actual value" do
+                actual = { :an_key => "dummy", :fixed => "fixed", :nested => { :anything_key => "bcdd0399-1cfe-4", :name => "foo", :trigger => "trigger" } }
+                expected = { :an_key => anything, :fixed => "fixed", :nested => { :anything_key => anything, :name => "foo", :trigger => "wrong" } }
+                diff = differ.diff(actual, expected)
+                expected_diff = dedent(<<-'EOD')
+                  |
+                  |@@ -1,4 +1,4 @@
+                  | :an_key => "dummy",
+                  | :fixed => "fixed",
+                  |-:nested => {:anything_key=>"bcdd0399-1cfe-4", :name=>"foo", :trigger=>"wrong"},
```
pirj, line 588: I’m in the fence about this change. For nested arrays, doesn’t it make more sense to indicate that a certain key was `anything`?

lib/rspec/support/differ.rb
```diff
@@ -72,6 +76,46 @@ def initialize(opts={})
       end
 
     private
+      def run(actual, expected)
+        diff = ""
+
+        no_procs_no_numbers = lambda {|var1, var2| no_procs?(var1, var2) && no_numbers?(var1, var2)}
```
pirj, line 82: Is there a reason not to make this a method?

lib/rspec/support/differ.rb
```diff
@@ -72,6 +76,46 @@ def initialize(opts={})
...
+
+        if all_strings?(actual, expected)
+          diff = diff_as_string(coerce_to_string(actual), coerce_to_string(expected)) if any_multiline_strings?(actual, expected)
+        elsif (keys = all_keys_from_hash(expected)) && keys.any?
```
pirj, line 86: What if we check no_procs_no_numbers here, and the other check goes inside?

spec/rspec/support/differ_spec.rb
KarlHeitmann: I don't have a strong opinion for nested arrays. Displaying that a certain key was anything for _nested_ arrays was the original behavior.

lib/rspec/support/differ.rb
KarlHeitmann: I don't know, I don't have a strong opinion either here. Originally [I did it this way](https://github.com/rspec/rspec-support/pull/596/commits/d50d7429fe0ac20744477deba49a13d8989d6a3b#diff-df5537736876b545aac78c1d539a1836932f9a2d6c2170e9457490988b47a3f4L22):

```
            if no_procs?(actual, expected) && no_numbers?(actual, expected)
               diff = diff_as_object_with_anything(actual, expected)
             end
          elsif no_procs?(actual, expected) && no_numbers?(actual, expected)
             diff = diff_as_object(actual, expected)
```

But Rubocop complained about offending `Metrics::PerceivedComplexity` rule, in order to lower the perceived complexity score I made the lambda function to reduce the number of `&&` operators by 1, and that made the offense disappear.

I didn't made a method because I didn't know if it was overkill or not to create a method that only me will use it in only two places. But I don't have a strong opinion, I can write the method if you think it is better :)

lib/rspec/support/differ.rb
KarlHeitmann: Nice, that makes it easier to read! Thanks! I've submitted a new commit with your suggestion.

spec/rspec/support/differ_spec.rb
pirj: I’m still not too familiar with this addition. Do you yhink it is possible to keep the effect of collapsed “an_key” which I undoubtedly like, but make no changes to the “anything_key”’s value in expected of a nested array?

lib/rspec/support/differ.rb
pirj: Something broke specs, I hope it’s not because of this cosmetic suggestion. Please feel free to revert if this is the case

lib/rspec/support/differ.rb
pirj: I mean is it possible to extract ‘no_procs_no_numbers?’ to a method? That would even further reduce the complexity, right? And (I may be mistaken) reduce memory allocations on subsequent calls.

spec/rspec/support/differ_spec.rb
KarlHeitmann: Yea, sure. If you checkout the [first commit](https://github.com/rspec/rspec-support/pull/596/commits/8c811465f38579ee72060a68656bcc306b7b2b1b) of this PR, you'll be on a state where there is no recursion, thus no further modifications on nested objects.

I have no problem at all to revert to that state. I don't have a strong opinion about morphing `anything` fuzzy matcher in nested objects. After all, when a nested object differs you get the entire object highlighted in red anyways, so I understand morphing deep nested values (like the one you are unfamiliar with: `:anything_key`) may not be worth to do, and it may be better to only keep the effect of collapsed first level keys like `"an_key"`.

But I don't know the answer to this, just want to give you options so you can choose a good answer. You have a lot more experience than I in this project :)

KarlHeitmann:
Hey @pirj , what if we just stick with [the first commit](https://github.com/rspec/rspec-support/pull/596/commits/8c811465f38579ee72060a68656bcc306b7b2b1b) and improve it with your great suggestion [here](https://github.com/rspec/rspec-support/pull/596/commits/47cf1b286817d4f2f9e2734ac5cacb31a3c2ed9e#r1607733563)?

The first commit is a simpler solution that implements the behavior you mentioned [here](https://github.com/rspec/rspec-support/pull/596#discussion_r1608665469), and I think that behavior although it is simpler is still a good improvement on `rspec-support` differ. `RSpec::Support::Differ` already displays nested object diffs in a single line, so the enhancement provided by my suggestion on this PR are marginal in such cases, it may not be worth to implement them.

I recall having seen the broken tests[ you mentioned](https://github.com/rspec/rspec-support/pull/596#discussion_r1608667126) while I was working on my last commits, but I am unable to reproduce the broken tests locally. Maybe it is better to simplify the PR, I can even open a new & cleaner PR with the first commit enhanced by your suggestion.

pirj:
Completely agree, @KarlHeitmann. This pr will be kept if we decide to also include diffing for lines.

KarlHeitmann:
Ready, I've just submitted #599 with the simplified version of this PR.

JonRowe:
:wave: Do you want to rebase this taking #599 into account? If you'd prefer to close this thats also fine

KarlHeitmann:
Hey @JonRowe !

To be honest, I don't know what to do here...

I couldn't rebase the branch, because the merged #599 PR turned out to be completely different to what I initially wrote on the first commit of this branch.

Nevertheless, on my local machine I completely rewrote the current branch taking #599 and had success. The changes I made are simpler than what I did on my initial approach on the current PR #596 .

I could force-push my local branch into my repo so you can see the changes. _But I don't know if it is worth to do._  Because the diff message for nested hashes is already confusing. I mean, if I made the changes to hide any value that matches the `anything` fuzzy matcher on nested hashes.... it will not make a significant difference, because I would still see a biiiiiiiig single line diff for a hash object anyways 🤷. I don't know if I'm making myself understood correctly.

OMHO, I think I can close this PR (or keep it "on hold") until `RSpec::Support::Differ` class can handle diffing nested hashes in a readable and pretty way. Once that task is done, continue with the current PR to hide `anything` matches on nested hashes.

But I think it is your call, I can force push my local branch here or close this PR. Also, I can try to fix `RSpec::Support::Differ` to display nested hashes diff in a prettier way.

pirj:
@KarlHeitmann what is your plan of improving the diff from the description?
The first, non-nested, part is done, right?

As I’m looking at before/after, my preference is to keep things as is. There are certainly cases where diffs can be improved. But any improvement is welcome. And I kindly appreciate your thoughtful and thorough approach!

Please feel free to force-push the rebased branch here.

JonRowe:
I'm going to close this as is, if / when the differ gets improved over in RSpec Support and you'd like to re-open this as a new improvement then feel free to do so :)

PR #597, opened 2024-05-09
Show a diff for comparisons of arrays of one-line strings

nevinera:
A long time ago, a fellow notice a regression in the [behavior of diffing](https://github.com/rspec/rspec/issues/78). Roughly speaking, `expect(["foo"]).to eq(["foo", "baz"])` doesn't print a "Diff" section of output. But `expect(["foo"]).to eq(["foo", "baz\n"])` _does_. And that seems like an odd distinction to make! To the code mines!

I had to trace this across the move from rspec-expectations into rspec-support, but it was originally introduced [here](https://github.com/rspec/rspec-expectations/commit/044b0a6ff1), which was motivated by [this issue](https://github.com/rspec/rspec-expectations/issues/115). That seems like a reasonable goal, and we do actually have some tests enforcing it already (though they're a little entangled with the encoding specs - I can add another test specific to this purpose if you like). The effect it had on diffing _arrays_ of strings seems incidental though, a result of the way we flatten the arrays to test them, rather than checking them individually.

So! Break that check out ahead, and make it a bit clearer - add `pointless_diff?`, which tests that both arguments are Strings, and that neither string is multiline; for that case, skip adding the diff like before, but drop the 'multiline' test that _was_ skipping.

### DIFF
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -13,11 +13,9 @@ module RSpec
       def diff(actual, expected)
         diff = ""
 
-        unless actual.nil? || expected.nil?
+        unless actual.nil? || expected.nil? || pointless_diff?(actual, expected)
           if all_strings?(actual, expected)
-            if any_multiline_strings?(actual, expected)
-              diff = diff_as_string(coerce_to_string(actual), coerce_to_string(expected))
-            end
+            diff = diff_as_string(coerce_to_string(actual), coerce_to_string(expected))
           elsif no_procs?(actual, expected) && no_numbers?(actual, expected)
             diff = diff_as_object(actual, expected)
           end
@@ -81,8 +79,8 @@ module RSpec
         safely_flatten(args).all? { |a| String === a }
       end
 
-      def any_multiline_strings?(*args)
-        all_strings?(*args) && safely_flatten(args).any? { |a| multiline?(a) }
+      def pointless_diff?(actual, expected)
+        String === actual && String === expected && !multiline?(actual) && !multiline?(expected)
       end
 
       def no_numbers?(*args)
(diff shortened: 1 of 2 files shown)


nevinera:
Ah, and _now_ I think I get to learn why the mono-repo is such a desirous outcome :-)

nevinera:
> I'm going to see if I can accommodate that intent easily.

Narrator: "He could not."

It's essentially doing _exactly_ the thing we were trying to change; diffing two arrays of two single-line strings. So - the options are (a) update the specs to reflect a new intent, (b) cancel this plan, or (c)switch plans - the inconsistency could also be resolved by just not diffing arrays of strings _at all_.

I favor (a), despite the extra PRs it'll involve (go, monorepo!) - the output it produces looks like these, which seems sufficiently useful to me:

```
     Failure/Error: d.foo("argA", "argB", "ARGTHREE", "arg2")

       #<Double "double"> received :foo with unexpected arguments
         expected: ("arg1", "arg2", "arg3", "arg4")
              got: ("argA", "argB", "ARGTHREE", "arg2")
       Diff:

       @@ -1,5 +1,5 @@
       -arg1
       +argA
       +argB
       +ARGTHREE
        arg2
       -arg3
       -arg4


    Failure/Error: d.foo("arg1", "arg2", "ARGTHREE", "arg4")

       #<Double "double"> received :foo with unexpected arguments
         expected: ("arg1", "arg2", "arg3", "arg4")
              got: ("arg1", "arg2", "ARGTHREE", "arg4")
       Diff:
       @@ -1,5 +1,5 @@
        arg1
        arg2
       -arg3
       +ARGTHREE
        arg4
```

nevinera:
@pirj/@JonRowe , does that plan seem appropriate?

pirj:
I’m all for the plan a!

JonRowe:
@nevinera I'm closing this as it is a draft as part of the monorepo migration, if you'd like to try it on the monorepo please do!

PR #599, opened 2024-05-25
Fix diff output when a fuzzy finder anything is inside an expected hash

KarlHeitmann:
This is a simplified version of PR #596 , it solves the issue #551 "Diff reports confusing output when used with "fuzzy" matchers like `anything` "

This PR will fix only `anything` values associated to top level keys of a hash. It will not work with nested hashes.

Example output *BEFORE* changes in this PR:

```
@@ -1,4 +1,4 @@
-:anything_key => anything,
+:anything_key => "bcdd0399-1cfe-4de1-a481-ca6b17d41ed8",
 :fixed => "fixed",
-:trigger => "wrong",
+:trigger => "trigger",
```

Example output *AFTER* changes in this PR:

```
@@ -1,4 +1,4 @@
 :anything_key => anything,
 :fixed => "fixed",
-:trigger => "wrong",
+:trigger => "trigger",
```

### DIFF
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -58,8 +58,8 @@ module RSpec
       end
       # rubocop:enable Metrics/MethodLength
 
-      def diff_hashes_as_object(actual, expected)
-        if defined?(RSpec::Mocks::ArgumentMatchers::AnyArgMatcher)
+      if defined?(RSpec::Mocks::ArgumentMatchers::AnyArgMatcher)
+        def diff_hashes_as_object(actual, expected)
           expected_to_diff =
             expected.reduce({}) do |hash, (key, value)|
               if RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === value
@@ -71,7 +71,9 @@ module RSpec
             end
 
           diff_as_object(actual, expected_to_diff)
-        else
+        end
+      else
+        def diff_hashes_as_object(actual, expected)
           diff_as_object(actual, expected)
         end
       end
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -19,7 +19,7 @@ module RSpec
               diff = diff_as_string(coerce_to_string(actual), coerce_to_string(expected))
             end
           elsif no_procs_and_no_numbers?(actual, expected)
-            if Hash === expected && hash_with_anything?(expected)
+            if (RUBY_VERSION.to_f > 1.8) && hash_with_anything?(expected)
               diff = diff_as_object_with_anything(actual, expected)
             else
               diff = diff_as_object(actual, expected)
@@ -85,7 +85,7 @@ module RSpec
     private
 
       def hash_with_anything?(arg)
-        safely_flatten(arg).any? { |a| RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === a }
+        Hash === arg && safely_flatten(arg).any? { |a| RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === a }
       end
 
       def no_procs_and_no_numbers?(*args)
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -59,10 +59,25 @@ module RSpec
       # rubocop:enable Metrics/MethodLength
 
       def diff_hashes_as_object(actual, expected)
-        expected.select { |_, v| RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === v }.each_key do |k|
-          expected[k] = actual[k]
+        if defined?(RSpec::Mocks::ArgumentMatchers::AnyArgMatcher)
+          anything_hash = expected.select { |_, v| RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === v }
+
+          anything_hash.each_key do |k|
+            expected[k] = actual[k]
+          end
+
+          diff_string = diff_as_object(actual, expected)
+
+          if defined?(RSpec::Mocks::ArgumentMatchers::AnyArgMatcher)
+            anything_hash.each do |k, v|
+              expected[k] = v
+            end
+          end
+
+          diff_string
+        else
+          diff_as_object(actual, expected)
         end
-        diff_as_object(actual, expected)
       end
 
       def diff_as_object(actual, expected)
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -60,17 +60,16 @@ module RSpec
 
       if defined?(RSpec::Mocks::ArgumentMatchers::AnyArgMatcher)
         def diff_hashes_as_object(actual, expected)
-          expected_to_diff =
-            expected.reduce({}) do |hash, (key, value)|
-              if RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === value
-                hash[key] = actual[key]
-              else
+          actual_to_diff =
+            actual.reduce({}) do |hash, (key, value)|
+              if RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === expected[key]
                 hash[key] = expected[key]
+              else
+                hash[key] = actual[key]
               end
               hash
             end
-
-          diff_as_object(actual, expected_to_diff)
+          diff_as_object(actual_to_diff, expected)
         end
       else
         def diff_hashes_as_object(actual, expected)
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -18,8 +18,12 @@ module RSpec
             if any_multiline_strings?(actual, expected)
               diff = diff_as_string(coerce_to_string(actual), coerce_to_string(expected))
             end
-          elsif no_procs?(actual, expected) && no_numbers?(actual, expected)
-            diff = diff_as_object(actual, expected)
+          elsif no_procs_and_no_numbers?(actual, expected)
+            if Hash === expected && hash_with_anything?(expected)
+              diff = diff_as_object_with_anything(actual, expected)
+            else
+              diff = diff_as_object(actual, expected)
+            end
           end
         end
 
@@ -56,6 +60,13 @@ module RSpec
       end
       # rubocop:enable Metrics/MethodLength
 
+      def diff_as_object_with_anything(actual, expected)
+        expected.select { |_, v| RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === v }.each_key do |k|
+          expected[k] = actual[k]
+        end
+        diff_as_object(actual, expected)
+      end
+
       def diff_as_object(actual, expected)
         actual_as_string = object_to_string(actual)
         expected_as_string = object_to_string(expected)
@@ -72,6 +83,14 @@ module RSpec
       end
 
     private
+
+      def hash_with_anything?(arg)
+        safely_flatten(arg).any? { |a| RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === a }
+      end
+
+      def no_procs_and_no_numbers?(*args)
+        no_procs?(args) && no_numbers?(args)
+      end
 
       def no_procs?(*args)
         safely_flatten(args).none? { |a| Proc === a }
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -61,7 +61,7 @@ module RSpec
       if defined?(RSpec::Mocks::ArgumentMatchers::AnyArgMatcher)
         def diff_hashes_as_object(actual, expected)
           actual_to_diff =
-            actual.reduce({}) do |hash, (key, value)|
+            actual.keys.reduce({}) do |hash, key|
               if RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === expected[key]
                 hash[key] = expected[key]
               else
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -92,7 +92,7 @@ module RSpec
         no_procs?(args) && no_numbers?(args)
       end
 
-      def no_procs?(*args)
+      def no_procs?(args)
         safely_flatten(args).none? { |a| Proc === a }
       end
 
@@ -104,7 +104,7 @@ module RSpec
         all_strings?(*args) && safely_flatten(args).any? { |a| multiline?(a) }
       end
 
-      def no_numbers?(*args)
+      def no_numbers?(args)
         safely_flatten(args).none? { |a| Numeric === a }
       end
 
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -68,10 +68,8 @@ module RSpec
 
           diff_string = diff_as_object(actual, expected)
 
-          if defined?(RSpec::Mocks::ArgumentMatchers::AnyArgMatcher)
-            anything_hash.each do |k, v|
-              expected[k] = v
-            end
+          anything_hash.each do |k, v|
+            expected[k] = v
           end
 
           diff_string
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -18,12 +18,10 @@ module RSpec
             if any_multiline_strings?(actual, expected)
               diff = diff_as_string(coerce_to_string(actual), coerce_to_string(expected))
             end
-          elsif no_procs_and_no_numbers?(actual, expected)
-            if (RUBY_VERSION.to_f > 1.8) && hash_with_anything?(expected)
-              diff = diff_as_object_with_anything(actual, expected)
-            else
-              diff = diff_as_object(actual, expected)
-            end
+          elsif all_hashes?(actual, expected)
+            diff = diff_hashes_as_object(actual, expected)
+          elsif no_procs?(actual, expected) && no_numbers?(actual, expected)
+            diff = diff_as_object(actual, expected)
           end
         end
 
@@ -60,7 +58,7 @@ module RSpec
       end
       # rubocop:enable Metrics/MethodLength
 
-      def diff_as_object_with_anything(actual, expected)
+      def diff_hashes_as_object(actual, expected)
         expected.select { |_, v| RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === v }.each_key do |k|
           expected[k] = actual[k]
         end
@@ -84,16 +82,12 @@ module RSpec
 
     private
 
-      def hash_with_anything?(arg)
-        Hash === arg && safely_flatten(arg).any? { |a| RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === a }
-      end
-
-      def no_procs_and_no_numbers?(*args)
-        no_procs?(args) && no_numbers?(args)
-      end
-
-      def no_procs?(args)
+      def no_procs?(*args)
         safely_flatten(args).none? { |a| Proc === a }
+      end
+
+      def all_hashes?(actual, expected)
+        defined?(RSpec::Mocks::ArgumentMatchers::AnyArgMatcher) && (Hash === actual) && (Hash === expected)
       end
 
       def all_strings?(*args)
@@ -104,7 +98,7 @@ module RSpec
         all_strings?(*args) && safely_flatten(args).any? { |a| multiline?(a) }
       end
 
-      def no_numbers?(args)
+      def no_numbers?(*args)
         safely_flatten(args).none? { |a| Numeric === a }
       end
 
--- a/lib/rspec/support/differ.rb
+++ b/lib/rspec/support/differ.rb
@@ -60,19 +60,17 @@ module RSpec
 
       def diff_hashes_as_object(actual, expected)
         if defined?(RSpec::Mocks::ArgumentMatchers::AnyArgMatcher)
-          anything_hash = expected.select { |_, v| RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === v }
+          expected_to_diff =
+            expected.reduce({}) do |hash, (key, value)|
+              if RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === value
+                hash[key] = actual[key]
+              else
+                hash[key] = expected[key]
+              end
+              hash
+            end
 
-          anything_hash.each_key do |k|
-            expected[k] = actual[k]
-          end
-
-          diff_string = diff_as_object(actual, expected)
-
-          anything_hash.each do |k, v|
-            expected[k] = v
-          end
-
-          diff_string
+          diff_as_object(actual, expected_to_diff)
         else
           diff_as_object(actual, expected)
         end
@@ -100,7 +98,7 @@ module RSpec
       end
 
       def all_hashes?(actual, expected)
-        defined?(RSpec::Mocks::ArgumentMatchers::AnyArgMatcher) && (Hash === actual) && (Hash === expected)
+        (Hash === actual) && (Hash === expected)
       end
 
       def all_strings?(*args)
--- a/spec/rspec/support/differ_spec.rb
+++ b/spec/rspec/support/differ_spec.rb
@@ -6,7 +6,7 @@ require 'rspec/support/spec/string_matcher'
 
 module RSpec
   module Support
-    RSpec.describe Differ do
+    RSpec.describe "Differ" do
       include Spec::DiffHelpers
 
       describe '#diff' do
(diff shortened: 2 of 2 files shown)


pirj:
> undefined method `flatten' for #<Hash

in 1.8.7

Looks great otherwise, thank you!

pirj:
For the elevator pitch, may I kindly as you to add before and after example outputs to the pr description?

KarlHeitmann:
Ready!

- I've just added the before/after example outputs on the PR description.

- The error `undefined method  flatten for #<Hash>` is really weird. I am not calling that method anywhere on the `_spec.rb` file (the file that is complaining), and the application file `differ.rb` [calls flatten on line 135](https://github.com/rspec/rspec-support/blob/8cbf7f9aecbefacd74da16ec2e713358ff72925b/lib/rspec/support/differ.rb#L135). But [I have not touched that line on my branch](https://github.com/KarlHeitmann/rspec-support/blob/fix_diff_fuzzy_matcher_anything_v2/lib/rspec/support/differ.rb#L154). I made a workaround to bypass the new code if you are using ruby version 1.8.7. That should be enough to solve the issue...

lib/rspec/support/differ.rb
```diff
@@ -73,6 +84,14 @@ def initialize(opts={})
 
     private
 
+      def hash_with_anything?(arg)
+        Hash === arg && safely_flatten(arg).any? { |a| RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === a }
```
pirj, line 88: It seems this is due to this. We call safely_flatten with a hash arg, and it calls flatten on it.
This will break 1.8.7. We can’t afford this. Even though I’m a proponent of soft-deprecating older Rubies in our code, this doesn’t include breaking hypothetical suites that might still exist.

Can the same be achieved somehow differently? Like taking `.values` from the hash?

pirj: Please disregard my note regarding breaking 1.8.7, I’ve just noticed that you have a check for this.

Still, does it make sense to flatten? We could just check the too level of the hash, right?

lib/rspec/support/differ.rb
```diff
@@ -73,6 +84,14 @@ def initialize(opts={})
...
+      end
+
+      def no_procs_and_no_numbers?(*args)
```
pirj, line 91: I’m also struggling to understand how this works.
 We do pass expected and actual there, and they end up wrapped in an array when flatten is called on it?

KarlHeitmann: Indeed... that's weird... What if I remove the star operator from `#no_procs` and `no_numbers` method?

The purpose of this `no_procs_and_no_numbers?` method is to reduce the number of `&&` operators used in the `#diff` method, thus appeasing rubocop by removing the `Metrics::PerceivedComplexity` offense.

KarlHeitmann:
I've just simplified `no_procs_and_no_numbers?` in a new commit by removing the star operator.

I don't know if that's the best solution, the downside is it modifies the original `no_procs?` and `no_numbers?` input argument. Usually I don't like to alter the original code that existed before doing my changes.

Other solutions I could have  taken that didn't need to alter the `no_procs?` and `no_numbers?` methods are:

- Alternative 2: copy paste the logic of `no_procs?` and `no_numbers?` and joins the expression with `&&` operator
```ruby
      def no_procs_and_no_numbers?(*args)
        safely_flatten(args).none? { |a| Proc === a } && safely_flatten(args).none? { |a| Numeric === a }
      end
```

- Alternative 3: an optimization of Alternative 2. This one will have better performance
```ruby
      def no_procs_and_no_numbers?(*args)
        safely_flatten(args).none? { |a| Proc === a || Numeric === a }
      end
```

Both alternatives makes all test pass on my local machine.

lib/rspec/support/differ.rb
```diff
@@ -73,7 +84,15 @@ def initialize(opts={})
 
     private
 
-      def no_procs?(*args)
+      def hash_with_anything?(arg)
+        Hash === arg && safely_flatten(arg).any? { |a| RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === a }
+      end
+
+      def no_procs_and_no_numbers?(*args)
+        no_procs?(args) && no_numbers?(args)
```
pirj, line 92: args here will be an array of two elements?
How does it work to flatten an array of two hashes?

Can we check no_procs?(expected) && no_procs?(actual) && no_numbers? …. 

Each if those methods runs the (potentially expensive?) safe_flatten. I’d prefer this to be done just once (or, better - never!)

Can’t we just take top-level keys of each of those arrays without flattening recursively?

This would work on 1.8.7
No tricks with has/array structures
Early exit if there is a proc/number
No extra checks in nested values that we don’t care about
Less code
No splatting/desplatting

KarlHeitmann:
UPDATE: yeah... what you are mentioning here is a good idea: get rid of the `safely_flatten` by just taking the top-level keys of each of those elements: `actual` & `expected`. I am doing some benchmarks because I have not found the complexity of Hash methods like `.none?` or `.any?` on the ruby [hash documentation](https://ruby-doc.org/3.2.2/Hash.html) or ruby [enumerable documentation](https://ruby-doc.org/core-3.0.2/Enumerable.html). I hope I'll have a simple proposal ready in the next days to submit it here :)

KarlHeitmann:
I realized the change is simpler than I thought. I just introduced a new commit that reverts the `no_procs_and_no_numbers?` method and restores `no_procs?` and `no_numbers?` to its original state. This new commit introduces `Differ#all_hashes?(actual, expected)` private method that will check if `actual` and `expected` are Hashes. In such cases, it will call the method that checks the presence of `RSpec::Mocks::ArgumentMatchers::AnyArgMatcher` fuzzy matcher as value on the expected hash, then will mutate its `anything` value and will fold that diff on the output string.

I am doing this instead of using `safely_flatten` because the behavior I'm proposing on this PR is a specific case scenario: the user has given an `actual` and `expected` vars that both are `Hash` instances, and the `expected` var can contain an `anything` fuzzy matcher, so the output diff will contain less noise.

`safely_flatten` was introduced in [fc01e3e5b4e711a471dde0dd65d5e528439e9fdf](https://github.com/rspec/rspec-support/commit/fc01e3e5b4e711a471dde0dd65d5e528439e9fdf), and before this commit, methods such as `def no_procs?(*args)` were written like this:

```ruby
def no_procs?(*args)
  args.flatten.none? { |a| Proc === a }
end
```

But that method was only called once in that old file. You can check it out with this command:
```
git show fc01e3e5~:lib/rspec/support/differ.rb
```

This leads me to think that before using `safely_flatten`, calling `args.flatten` inside `no_procs?(*args)` was just a fancy way to check the `actual` and `expected` vars were not `Proc` objects; instead of using something like this:

```ruby
def no_procs?(actual, expected)
  (Proc != actual) && (Proc != expected)
end
``` 
The fancy way is more readable, and it has the benefit if `actual` or `expected` are `Array` objects, they can be recursively flattened and you could `diff_as_object` objects nested deep down in `actual` & `expected` arrays, as long as these nested objects are not `Proc`s. Pretty cool, huh? The problem with this approach is that you _could_ enter into an infinite recursion, that's why later on `safely_flatten` method was introduced.

But again, the code changes I'm proposing are narrowed down to a specific case scenario, where the user wants to diff two hashes in which one of them an `anything` fuzzy matcher is being used as a value of a top-level key of the `expected` var. No need to search recursively. Recursion should be addressed in my previous PR #596

KarlHeitmann:
I've just submitted another commit, because I noticed another issue: [on this line](https://github.com/KarlHeitmann/rspec-support/blob/c3e9ea926463829a048a131f7b85fef16c186ef5/lib/rspec/support/differ.rb#L63), the expected var *mutates*. So, if you provide `Differ` with a `expected` hash that contains an `anything` fuzzy matcher as a value in one of its top-level keys, after the `differ.diff` call the `expected` var will no longer contain the fuzzy matcher: it will contain the corresponding `actual` var value.

Hopefully the spec I've added will explain what the issue better than how I am explaining it here. Here is the code snippet:

```ruby
it "checks the 'expected' var continues having the 'anything' fuzzy matcher, it has not mutated" do
  actual = { :fixed => "fixed", :trigger => "trigger", :anything_key => "bcdd0399-1cfe-4de1-a481-ca6b17d41ed8" }
  expected = { :fixed => "fixed", :trigger => "wrong", :anything_key => anything }
  differ.diff(actual, expected)
  expect(expected).to eq({ :fixed => "fixed", :trigger => "wrong", :anything_key => anything })
end
```

Here is the output before the application code changes (failing test)

image

```
     Failure/Error: DEFAULT_FAILURE_NOTIFIER = lambda { |failure, _opts| raise failure }
     
       expected: {:anything_key=>anything, :fixed=>"fixed", :trigger=>"wrong"}
            got: {:anything_key=>"bcdd0399-1cfe-4de1-a481-ca6b17d41ed8", :fixed=>"fixed", :trigger=>"wrong"}
     
       (compared using ==)

```

Here is the output AFTER the application code changes (test succeeds)

image

pirj:
Nice for nd. Can this be the root cause of this getting its way to the diff in the first place?
```
1,4 @@
-:anything_key => anything,
+:anything_key => "bcdd0399-1cfe-4de1-a481-ca6b17d41ed8",
```

lib/rspec/support/differ.rb
```diff
@@ -56,6 +58,28 @@ def diff_as_string(actual, expected)
...
+        if defined?(RSpec::Mocks::ArgumentMatchers::AnyArgMatcher)
+          anything_hash = expected.select { |_, v| RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === v }
+
+          anything_hash.each_key do |k|
+            expected[k] = actual[k]
+          end
+
+          diff_string = diff_as_object(actual, expected)
+
+          if defined?(RSpec::Mocks::ArgumentMatchers::AnyArgMatcher)
```
pirj, line 71: Is this necessary? Looks like the same check on line 62

lib/rspec/support/differ.rb
```diff
@@ -56,6 +58,28 @@ def diff_as_string(actual, expected)
       end
       # rubocop:enable Metrics/MethodLength
 
+      def diff_hashes_as_object(actual, expected)
+        if defined?(RSpec::Mocks::ArgumentMatchers::AnyArgMatcher)
```
pirj, line 62: We only call this from all_hashes?, and it already has this check

Potentially, this method can be mistakenly called from somewhere else causing errors for those not using rspec-mocks, but i’m less worried about this at this point.

lib/rspec/support/differ.rb
```diff
@@ -77,6 +101,10 @@ def no_procs?(*args)
         safely_flatten(args).none? { |a| Proc === a }
```
pirj, line 101: What I really liked about your previous changes was the potential to get rid of the recursive check and the safely_flatten method.
No pressure, as this is not the goal of the pr, but it would be a nice bonus to fixing the issue. And a performance improvement in theory.

spec/rspec/support/differ_spec.rb
```diff
@@ -555,6 +555,30 @@ def inspect; "<BrokenObject>"; end
...
+
+        describe "fuzzy matcher anything" do
+          it "outputs only key value pair that triggered diff, anything_key should absorb actual value" do
+            actual = { :fixed => "fixed", :trigger => "trigger", :anything_key => "bcdd0399-1cfe-4de1-a481-ca6b17d41ed8" }
+            expected = { :fixed => "fixed", :trigger => "wrong", :anything_key => anything }
+            diff = differ.diff(actual, expected)
+            expected_diff = dedent(<<-'EOD')
+              |
+              |@@ -1,4 +1,4 @@
+              | :anything_key => "bcdd0399-1cfe-4de1-a481-ca6b17d41ed8",
```
pirj, line 567: Now when I look at this, this is indeed and improvement to `-anything_key +anything_key` we had.
But do we need a potentially huge value here? An uuid is fine, but what about a 5kb json-like thing?

Most certainly if we’ve `anything`ed it in the test, we don’t care what’s inside, including the diff?

@JonRowe does this make sense?

lib/rspec/support/differ.rb
KarlHeitmann: Yes, it is currently called only under `if all_hashes?(actual, expected)` condition, **and** `all_hashes?` is **also** checking if the user is opting out `rspec-mocks`. This check in line 62 is redundant if we assume this method *will only be called inside `all_hashes?` condition* . But it is there in case someone mistakenly call the method from somewhere else, so the program won't break if `rspec-mocks` is opted-out.

If someone opts-out `rspec-mock`, `diff_hashes_as_object` should have the exact same behavior as `diff_as_object`. That's the purpose of this redundant check.

Personally, I like redundancy. But I understand sometimes it may be overkill or unnecessary... maybe defining `diff_hashes_as_object` as private is good enough? I don't have a strong opinion here.

KarlHeitmann:
Yes, the cause of:
```
1,4 @@
-:anything_key => anything,
+:anything_key => "bcdd0399-1cfe-4de1-a481-ca6b17d41ed8",
```

is the mutation there: First I take the `expected` var, I ask it if it is a `Hash` instance, and if it is a `Hash`, I'll select its values on the top-level keys that are `anything` and mutate them into their corresponding `actual` value. That's the core idea to fold all that noise. On the first PR #596  I applied that to all nested hashes recursively... but that's not very useful because the current differ `RSpec::Support::Differ` shows a single red line change for any nested hash... it is not very useful and I don't think it is worth doing that (side note: although it is proven possible to do on #596 ).

Before submitting the [latest commit c3e9ea926463829a048a131f7b85fef16c186ef5](https://github.com/rspec/rspec-support/pull/599/commits/c3e9ea926463829a048a131f7b85fef16c186ef5) , I thought the mutation would occur inside the `Differ` class, but will not "bubble up" to the test scope... but I was wrong, that commit fixes that issue.

lib/rspec/support/differ.rb
KarlHeitmann: Ooops... not necessary, I missed erasing that line. I've just submitted a commit with the removal.

spec/rspec/support/differ_spec.rb
KarlHeitmann: Are you concerned by the case a 5kb json-like thing would be copied into the `expected -> anything` value? that will definitively be a performance issue. Would it be worth doing the change? Is there a way we can make sure the 5kb json-like thing will be a shallow copy?  Maybe a shallow copy will mitigate the performance issue.

KarlHeitmann: Maybe a shallow copy will not impact memory ram consumption. But what about CPU time? does `diff_as_object` iterates the `actual` and `expected` vars regardless of their `.object_id`? or it checks the `object_id` and if they match it will cease any iteration? 

Another question: `rspec-support` module should be allowed to mutate `expected` vars? regardless they will be later on return them to their original state? My proposed changes on this PR are a hacky way to fold noise when the developers use `anything` fuzzy matcher on their hash, I understand this may not be allowed to do here.

KarlHeitmann: Apparently, a 5kb json-like thing will impact CPU time ONLY if the matcher fails.

```ruby
  def diff_as_object(actual, expected)
    actual_as_string = object_to_string(actual) # STEP 1
    expected_as_string = object_to_string(expected) # STEP 2
    diff_as_string(actual_as_string, expected_as_string) # STEP 3
  end
```

STEP 1 and STEP 2 will get the string representation of 5kb json-like thing in two different strings. STEP 3 will perform the comparison of both strings. This will be an expensive operation because both strings will be huge.  In contrast, without the changes on this PR, comparing the string representation of this 5kb json-like thing with `anything` fuzzy matcher would have been easy-peasy

AFAIK, `RSpec::Support::Differ` is called when an [expectation matcher fails ](https://github.com/rspec/rspec-expectations/blob/50c3ebeed61671c369e9c106050daed1794cd28f/lib/rspec/expectations/handler.rb#L56)AND the expectation [matcher responds to diffable?](https://github.com/rspec/rspec-expectations/blob/50c3ebeed61671c369e9c106050daed1794cd28f/lib/rspec/expectations/handler.rb#L38). If the matcher does not responds to diffable, `expected` and `actual` vars are assigned `nil` and `RSpec::Support::Differ` will not perform any action at all.

My assessment is this will impact performance only when the test fails and the developer is using a matcher that produces an output-diff. But this is my assessment about performance, I still don't know the answer to: should `rspec-support` module be allowed to mutate `expected` vars in the developer test bench (ie, his `_spec.rb` files)? or `rspec-support` should remain a "pure" read-only module?

JonRowe: I think @pirj's concern was about larger diffs, but as things should be identical the differ will often not print anything at all

JonRowe (changes requested):
Sorry for the delayed review, I think this localised change makes sense as it should reduce diff confusion. I have some suggestions though.

KarlHeitmann:
Almost ready @JonRowe .  :warning:  :warning:  :warning: But there is a caveat :warning:  :warning:  :warning: 

- The [first commit  31f086e](https://github.com/rspec/rspec-support/pull/599/commits/31f086ec4df181e89e9bd1199c728971168d04d4) addresses your comments in which I could make safe changes.

- The [second commit dd7a4c7](https://github.com/rspec/rspec-support/pull/599/commits/dd7a4c7e169e17323671571b6e19c301e99da0fa) addresses [your comment](https://github.com/rspec/rspec-support/pull/599#discussion_r1639714262) related to conditionally define my method `diff_hashes_as_object` or just "alias" it to `diff_as_object`. But I discovered these changes can break your test suite by throwing a random false positive test failure.

If I run only that test file with the command:

```bash
bundle exec rspec ./spec/rspec/support/differ_spec.rb
```

A failure rises. Because the [RSpec.describe Differ](https://github.com/KarlHeitmann/rspec-support/blob/0dea76973799b5f9b1381722a0510a86dcbea1ea/spec/rspec/support/differ_spec.rb#L9) sentence :warning: :warning: :warning:  will load `RSpec::Support::Differ` _before_ `RSpec::Mocks::ArgumentMatchers::AnyArgMatcher` is required on RSpec :warning: :warning: :warning: . So, the test I introduced with the `anything` matcher will _always fail_ with the command above: because it will define the `alias` version of my method instead of defining my method that handles `anything` fuzzy matcher.

But if you run the full test suite with:

```bash
bundle exec rspec spec
```

Chances are all tests will pass, unless  `differ_spec.rb` file IS the 1st file to be run. I've ran the full test suite many many times (20+ and I got bored) and I was always lucky. I've introduced a `puts "FLAKEY TEST FUZZY MATCHER RUN"` sentence on the flaky test,  Below is an example screenshot with 4 trials... but if your CI CD has very bad luck, you may find some false positives.

image

I don't know how to fix this issue, is there a way to force the test suite to not run `differ_spec.rb` first? But if we do so, anyone that wants to add something to `RSpec::Support::Differ` will stumble upon this issue when running `rspec ./spec/rspec/support/differ_spec.rb`

I've tried to dig in and find _where_ `AnyArgMatcher` is loaded to try to force `differ_spec.rb` to load it before the `RSpec.describe Differ do` sentence. But I didn't had success. the furthest I went is here: [rspec-core/lib/rspec/core/dsl.rb#L43](https://github.com/rspec/rspec-core/blob/1e661db5c5b431c0ee88a383e8e3767f02dccbfe/lib/rspec/core/dsl.rb#L43) , on that line, I've put a `binding.break` before the class gets defined. I hope this screenshot can provide some more information:

image

What should we do about this?

lib/rspec/support/differ.rb
pirj: I can’t tell of the top of my head with certainty, but I can’t recall if nested hashes are (usefully) diffed.

I’d rather simplify the code if no spec fails.

pirj:
What if we quote it?
```ruby
RSpec.describe "Differ" do
```
it’s the way we recommend writing specs anyway. Would it solve the flakiness issue?


Alternatively. We autoload classes, and numerous places may trigger `AnyArgMatcher` to load.
If we have no intention to run specs when rspec-mocks are not loaded on purpose, then I don’t see any harm if you require the file from rspec-mocks that defines it explicitly at the bottom of this spec file.

KarlHeitmann:
That's awesome! `RSpec.describe "Differ" do` fixed the issue! I've run the single test file and the whole test suite and I had no problems at all.

If there are no more changes left to do, I can squash all the commits into a single one.

BTW, What's the difference between RSpec.describe the class and the name of the class as string?

pirj:
Nice! Let me have another quick look.

> RSpec.describe the class and the name of the class as string?

This autoloading (this is the first in my practice). And also probably `described_class` won’t work.

pirj:
> uninitialized constant RSpec::Support::HunkGenerator
     # ./spec/rspec/support/differ_spec.rb:169

on 2.7 🤔

spec/rspec/support/differ_spec.rb
pirj: Identical - yes. But if it’s neighbouring with the difference, like the ‘anything_key’ here on line 567, and it is big, i would prefer to see the literal “anything” to be printed rather than 5kb of json.

pirj (approved):
There are a few things with potential improvements in this code. But I don’t think they are related to this PR.

The change looks good. Thank you!

KarlHeitmann:
You're welcome! Is there anything else I can do for you here? @pirj  @JonRowe ? I can squash all the commits into a single one if you think this PR is OK.

BTW, in Ruby 2.7 build it is complaining about [spec/rspec/support/differ_spec.rb:169](https://github.com/rspec/rspec-support/blob/5334c64c77e3284ad417e3bea5805f5728d81022/spec/rspec/support/differ_spec.rb#L169)

```
  1) Differ #diff handles any encoding error that occurs with a helpful error message
     Failure/Error:
       expect(RSpec::Support::HunkGenerator).to receive(:new).
         and_raise(Encoding::CompatibilityError)

     NameError:
       uninitialized constant RSpec::Support::HunkGenerator
```

I don't know how can I reproduce this error on my local machine. I have not touched this line 169 before, is it a flaky test? or is it related to changing:

```ruby
RSpec.describe Differ do
```

to

```ruby
RSpec.describe "Differ" do
```

?

pirj:
2.7 passed on re-run. Must be order-dependent load order flakiness. Let’s see if it bugs us again. But I don’t think it should stop from merging.

Thanks again.
@JonRowe wdyt?

JonRowe:
Ugh I hate to say this at this late juncture, but I think @pirj's earlier point about large nested hashes has a point, if our point here is to make the diff smaller by making things equal, and the matcher is matching on `anything` then why aren't we doing the inverse of this? e.g. Creating a `actual_to_diff` and setting the ignored keys to "anything"

lib/rspec/support/differ.rb
```diff
@@ -56,6 +58,26 @@ def diff_as_string(actual, expected)
...
+            expected.reduce({}) do |hash, (key, value)|
+              if RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === value
+                hash[key] = actual[key]
+              else
+                hash[key] = expected[key]
+              end
+              hash
+            end
+
+          diff_as_object(actual, expected_to_diff)
```
JonRowe, line 73: If we inverted this, it would look like this:
```suggestion
          actual_to_diff =
            actual.reduce({}) do |hash, (key, value)|
              if RSpec::Mocks::ArgumentMatchers::AnyArgMatcher === expected(key)
                hash[key] = expected[key]
              else
                hash[key] = actual[key]
              end
              hash
            end

          diff_as_object(actual_to_diff, expected)
```

KarlHeitmann:
That's thinking outside of the box! I tried your suggestion and it works! I didn't thought about it... but using the `actual_to_diff` approach _is better_ than using the `expected_to_diff` approach :+1:

spec/rspec/support/differ_spec.rb
```diff
@@ -555,6 +555,31 @@ def inspect; "<BrokenObject>"; end
...
+
+        describe "fuzzy matcher anything" do
+          it "outputs only key value pair that triggered diff, anything_key should absorb actual value" do
+            actual = { :fixed => "fixed", :trigger => "trigger", :anything_key => "bcdd0399-1cfe-4de1-a481-ca6b17d41ed8" }
+            expected = { :fixed => "fixed", :trigger => "wrong", :anything_key => anything }
+            diff = differ.diff(actual, expected)
+            expected_diff = dedent(<<-'EOD')
+              |
+              |@@ -1,4 +1,4 @@
+              | :anything_key => anything,
```
pirj, line 567: 🤩

pirj (approved):
Just perfect now 🙌 Thank you!
And thanks @JonRowe for the smart suggestion.

pirj:
HunkGenerator again, flaky

JonRowe:
Thanks!

PR #606, opened 2024-07-12
Stash methods that get and set thread local variables to allow the user to mock them

### LINKED ISSUE #580: This is not thread local, it is in fact fiber local.
https://github.com/rspec/rspec-support/blob/6d33eaa48a761d616e23b81fc33b68a6e68ec296/lib/rspec/support.rb#L93

I suggest introducing an actual thread local variable, e.g.

```ruby
class Thread
  attr_accessor :rspec_state
end
```

### LINKED ISSUE #605: Mocking `Thread.current.thread_variable_get` causes stack overflow.
Mocking `Thread.current.thread_variable_get` used to work with rspec-support <= 3.12.0. With rspec-support >= 3.12.1, it causes a stack overflow.

```ruby
# frozen_string_literal: true

begin
  require "bundler/inline"
rescue LoadError => e
  $stderr.puts "Bundler version 1.10 or later is required. Please update your Bundler"
  raise e
end

gemfile(true) do
  source "https://rubygems.org"

  gem "rspec", "3.12.0"
  ##gem "rspec-support", "3.12.0" # works
  gem "rspec-support", "3.12.1" # raises SystemStackError
end

puts "Ruby version is: #{RUBY_VERSION}"
require 'rspec/autorun'

RSpec.describe do
  it do
    expect(Thread.current).to receive(:thread_variable_get).with(:test).once.and_return(true)
    expect(Thread.current.thread_variable_get(:test)).to eq true
  end
end
```

### Your environment
* Ruby version: 3.3.1
* rspec-support version: 3.12.1 or 3.13.0 or 3.13.1 or current main branch (42b18c4ae6850244ede747ec348bfe4b04fa8876)

### Analysis

There is infinite recursion between the following steps:

1. Internal RSpec code calls `RSpec::Support.thread_local_data`.
2. `RSpec::Support.thread_local_data` calls `Thread.current.thread_variable_get`.
3. Because `Thread.current.thread_variable_get` is mocked, RSpec code is called. See 1.

### Workaround

Add `allow(Thread.current).to receive(:thread_variable_get).with(:__rspec).and_call_original`.

The intention of the original code was to mock only `Thread.current.thread_variable_get(:test)` but not `Thread.current.thread_variable_get` called with any other arguments. However, I could not find a way to do that.

### Proposed solution

If some mocked method is called from internal RSpec code, it should do nothing but delegate to the original method.

manueljacob:
Fixes #605.
Alternative to #581 to fix #580

### DIFF
--- a/lib/rspec/support.rb
+++ b/lib/rspec/support.rb
@@ -90,14 +90,23 @@ module RSpec
       singleton_class.ancestors.find { |ancestor| !ancestor.equal?(singleton_class) }
     end
 
-    # A single thread local variable so we don't excessively pollute that namespace.
+    # Stash original methods to allow the user to mock them.
     if RUBY_VERSION.to_f >= 2
+      THREAD_VARIABLE_GET = Thread.instance_method(:thread_variable_get)
+      THREAD_VARIABLE_SET = Thread.instance_method(:thread_variable_set)
+    else
+      THREAD_VARIABLE_GET = Thread.instance_method(:[])
+      THREAD_VARIABLE_SET = Thread.instance_method(:[]=)
+    end
+
+    # A single thread local variable so we don't excessively pollute that namespace.
+    if RUBY_VERSION >= '2.7'
       def self.thread_local_data
-        Thread.current.thread_variable_get(:__rspec) || Thread.current.thread_variable_set(:__rspec, {})
+        THREAD_VARIABLE_GET.bind_call(Thread.current, :__rspec) || THREAD_VARIABLE_SET.bind_call(Thread.current, :__rspec, {})
       end
     else
       def self.thread_local_data
-        Thread.current[:__rspec] ||= {}
+        THREAD_VARIABLE_GET.bind(Thread.current).call(:__rspec) || THREAD_VARIABLE_SET.bind(Thread.current).call(:__rspec, {})
       end
     end
 
--- a/spec/rspec/support_spec.rb
+++ b/spec/rspec/support_spec.rb
@@ -204,6 +204,17 @@ module RSpec
           end.resume
         end
       end
+
+      it "works when Thread#thread_variable_get and Thread#thread_variable_set are mocked" do
+        expect(Thread.current).to receive(:thread_variable_set).with(:test, true).once.and_return(true)
+        expect(Thread.current).to receive(:thread_variable_get).with(:test).once.and_return(true)
+
+        Thread.current.thread_variable_set(:test, true)
+        expect(Thread.current.thread_variable_get(:test)).to eq true
+
+        RSpec::Support.thread_local_data[:__for_test] = :oh_hai
+        expect(RSpec::Support.thread_local_data[:__for_test]).to eq :oh_hai
+      end
     end
 
     describe "failure notification" do


pirj:
Can you please add a spec that would fail before your changes?

pirj:
@ioquatix we fixed a side issue of not being able to mock get_instance_variable. Does it still work for your uses?

lib/rspec/support.rb
```diff
@@ -90,15 +90,18 @@ def self.class_of(object)
...
     end
 
-    # A single thread local variable so we don't excessively pollute that namespace.
+    # Stash original methods to allow the user to mock them.
     if RUBY_VERSION.to_f >= 2
-      def self.thread_local_data
-        Thread.current.thread_variable_get(:__rspec) || Thread.current.thread_variable_set(:__rspec, {})
-      end
+      @thread_variable_get = Thread.instance_method(:thread_variable_get)
+      @thread_variable_set = Thread.instance_method(:thread_variable_set)
```
JonRowe, line 96: Typically we've used a constant for these sort of stashing.
```suggestion
      THREAD_SET = Thread.instance_method(:thread_variable_set)
```

JonRowe:
I would argue that you shouldn't really be mocking the internals of thread, its not something you own, and this implementation still wouldn't allow you to mock `Thread.current` which is more likely IMO that individual methods...

ioquatix:
@pirj I stopped using RSpec so I don't have any strong opinion about it. However, I still think it would be better to use `Thread.attr_accessor :rspec_state` or something like that, would have also avoided this issue...

manueljacob:
> Fixes #605. Alternative to #581 to fix #580

@pirj You added this in the PR description. But unless I’m missing something, this PR is neither an alternative to #581 nor could have fixed #580. It fixes an issue introduced by #581.

manueljacob:
> Can you please add a spec that would fail before your changes?

Done. Not sure how I could miss this.

lib/rspec/support.rb
manueljacob: > Typically we've used a constant for these sort of stashing.

Done.

pirj: Can you please make sure you’ve pushed it?

manueljacob: Oops, I forgot to stash (no pun intended!) the changes before amending and pushing.

ioquatix (commented):
LGTM, however have you measured the performance overhead?

manueljacob:
> LGTM, however have you measured the performance overhead?

I pushed some performance optimization (using `UnboundMethod#bind_call`) instead of separate `UnboundMethod#bind` and `Method#call`, and did some quick tests. I replaced stdout and stderr to avoid writing 1000000 dots to the terminal.

```ruby
require 'benchmark'
require 'rspec'

out = StringIO.new

puts Benchmark::CAPTION
puts(Benchmark.measure do
  RSpec.describe do
    1000000.times do
      it do
      end
    end
  end

  RSpec::Core::Runner.run([], out, out)
end)

out.seek(0)
puts out.readlines[-3..-2]
```

Before this PR:
```
      user     system      total        real
 85.016527   2.258209  87.274736 ( 87.424472)
Finished in 59.82 seconds (files took 27.66 seconds to load)
1000000 examples, 0 failures
ruby rspec_perf.rb  86,45s user 2,69s system 99% cpu 1:29,46 total
```

After this PR:
```
      user     system      total        real
 85.154865   2.142675  87.297540 ( 87.447185)
Finished in 59.88 seconds (files took 27.65 seconds to load)
1000000 examples, 0 failures
ruby rspec_perf.rb  86,53s user 2,60s system 99% cpu 1:29,47 total
```

From these results, it’s only very slightly slower. However, the test conditions were not perfect (light background activity, power saving enabled).

manueljacob:
`UnboundMethod#bind_call` was added in Ruby 2.7, so I re-added the `UnboundMethod#bind` / `Method#call` code for older Ruby versions. On Ruby 2.6, I observed a much larger slowdown of up to 10% on the above benchmark. However, variance is high on my current benchmarking setup.

To do some proper benchmarking, I should follow [that](https://github.com/google/benchmark/blob/main/docs/reducing_variance.md) and [that](https://llvm.org/docs/Benchmarking.html). However, currently I don’t have a suitable machine available for that.

ioquatix:
Nice work, I assume pre-bind_call is more than trivially slower?

manueljacob:
> Nice work, I assume pre-bind_call is more than trivially slower?

The up to 10% slowdown on Ruby 2.6 described in my previous comment was this PR compared to the main branch. However, I’m not sure how much of the slowdown comes from the lack of `bind_call` vs. `bind`/`call` generally being less optimized on old Ruby versions. As I observed a lot a variance on the benchmark results especially on Ruby 2.6, I’ll be careful with conclusions.

I can do more benchmarks as soon as I can free some machine to do them without background load (which could be a few days). How detailed should it be? Does it even make sense to benchmark on end-of-life Ruby versions?

ioquatix:
I would personally not implement a feature like this if the performance was significantly impacted.

There are many users of RSpec - probably millions of invocations per day. How many of them want even 1-2% slower execution because of this feature?

JonRowe:
I'm closing this in favour of #610, I think in light of the conversation here, @ioquatix's original suggestion of patching in an accessor is warranted, especially because if people are mocking thread methods they are likely to mock `Thread.current`

PR #610, opened 2024-07-17
Switch rspec local data to a thread accessor

### LINKED ISSUE #580: This is not thread local, it is in fact fiber local.

JonRowe:
Originally suggested to fix rspec/rspec-support#580 I turned this down because of waryness over patching thread, but given the conversation about mocking in rspec/rspec#103 and rspec/rspec-support#606 I think this is a better solution than capturing methods and other overly complicated solutions.

Additionally we patch in an alias for `Thread.current` to protect access to that when mocked.

### DIFF
--- a/lib/rspec/support.rb
+++ b/lib/rspec/support.rb
@@ -2,6 +2,10 @@
 
 class Thread
   attr_accessor :__rspec_local_data
+
+  class << self
+    alias __rspec_current_thread current
+  end
 end
 
 module RSpec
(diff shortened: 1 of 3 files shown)


lib/rspec/support.rb
```diff
@@ -1,5 +1,13 @@
 # frozen_string_literal: true
 
+class Thread
+  attr_accessor :__rspec_local_data
+
+  class << self
+    alias __current_for_rspec current
```
ioquatix, line 7: I don't have a strong opinion about it but I always try to have a "namespace" first, e.g. `__rspec_current` (`__rspec_whatever`).

ioquatix (commented):
LGTM.

JonRowe:
Migrated to rspec/rspec#99

PR #611, opened 2024-08-19
[wip] Rework spec for different JRuby implementation.

JonRowe:


pirj:
@nevinera jfyi

spec/rspec/support/spec/stderr_splitter_spec.rb
```diff
@@ -101,19 +101,19 @@
...
+      splitter.reopen(tempfile)
+      expect(splitter.to_io.path).to eq(tempfile.path)
     ensure
-      $stderr.reopen(cloned)
+      splitter.reopen(cloned)
       tempfile.close
       tempfile.unlink
     end
-    expect($stderr.to_io).not_to be_a(File)
+    expect(splitter.to_io.path).not_to eq tempfile.path
```
nevinera, line 117: While I think this is a valid test for the code in StderrSplitter, I don't think it exercises the case I was trying to rely on in https://github.com/rspec/rspec-expectations/pull/1460 anymore. Depending on whether we actually want to release that PR, that may be a good thing? I hadn't actually intended [this PR](https://github.com/rspec/rspec-support/pull/598) to be merged until we had consensus on that one.

spec/rspec/support/spec/stderr_splitter_spec.rb
```diff
@@ -101,26 +101,28 @@
...
 
     cloned = splitter.clone
-    expect(splitter.to_io).not_to be_a(File)
+    expect(splitter.to_io).to be_stderr
 
     tempfile = Tempfile.new("foo")
     begin
       splitter.reopen(tempfile)
-      expect(splitter.to_io).to be_a(File)
+      expect(splitter.to_io).to_not be_stderr
```
JonRowe, line 114: @nevinera Unless I'm mistaken this represents the same effect as what you were testing, I'm not sure if I will merge this yet but it at least works for some JRuby

nevinera: Ah, I think the part I didn't follow was the bottom method - I assumed `be_stderr` was calling a `stderr?` method I wasn't aware of on the stream/splitter >.<

I'm not confident about how consistent that `inspect` is, but I'll trust your plan :-)

